PR #17, opened 2015-06-23
Removes unused hangupsjs client instance. Fixes #16

### LINKED ISSUE #16: Creates cookies.json in hangupsjs module directory
When starting yakyak it creates a `cookies.json` file in `resources/app/node_modules/hangupsjs/cookies.json` which isn't needed because the cookies are stored in `app.getPath('userData')`.

I've created a [package for Arch Linux](https://aur4.archlinux.org/packages/yakyak/) where yakyak is installed globally. This means that the `resources/app/node_modules/hangupsjs` directory is not writable.

As a workaround I'm currently touching `resources/app/node_modules/hangupsjs/cookies.json` and setting the permissions to that it's writable by anybody during the installation. But it would be nice if I didn't have to do this.

BTW: You can now install the "yakyak" package from AUR on Arch Linux and Manjaro.
```sh
$ yaourt -Sa yakyak
```
or 
```sh
$ packer -S yakyak
```

maxkueng:
This fixes #16 . The main thread currently creates a hangupsjs client instance `client = new Client()` without the correct options which is overridden with the right instance a couple lines below. The first instance without the `cookiespath` causes that hangupsjs tries to write a potentially not writable directory (e.g. is installed globally).

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -9,8 +9,6 @@ clipboard = require 'clipboard'
 
 tmp.setGracefulCleanup()
 
-client = new Client()
-
 app = require 'app'
 BrowserWindow = require 'browser-window'
 


algesten:
good catch! thanks! and very cool with arch Linux install!

PR #18, opened 2015-06-24
Enable audible notifications

### LINKED ISSUE #8: sound notifications
in which case do we want it?

"Any way we can get sound notifications for new IMs? Other than that this is off to a great start! Keep it up!"

davidbanham:
Fixes #8

Why not just use the audio notification already build into node-notify?

### DIFF
--- a/src/ui/views/notifications.coffee
+++ b/src/ui/views/notifications.coffee
@@ -59,6 +59,7 @@ module.exports = (models) ->
             message: text
             wait: true
             sender: 'com.github.yakyak'
+            sound: true
         , (err, res) -> if res?.trim() == 'Activate'
           action 'appfocus'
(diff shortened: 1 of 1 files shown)


algesten:
I did see that option, but couldn't decipher what the doc about it meant:

    sound: true, // Only Notification Center or Windows Toasters

but at least we get some sounds for some people. right? that's what it means?

davidbanham:
I think it translates to: "This works on OS X and on Windows 8"

It leaves Linux and earlier Windows out in the cold, which sucks. So we should ideally come up with a solution that works on all target platforms.

I think that makes more sense as a PR for node-notify, though, rather than adding the functionality to yakyak.

algesten:
+1. go for it.

Issue #3, opened 2015-06-19
system-wide proxy setting ignored


davibe:
The app ignores system wide proxy settings. This is confirmed on Windows but I suspect it's the same on Mac.

davidbanham:
Ideally, proxy support needs to be added to hangupsjs. As a hack, looks like we can just set an environment variable and request will pick it up.

davidbanham:
This has had to be removed temporarily due to a bug in Electron. We had a working workaround for OSX, but it failed on Windows.

algesten:
can win 7 be detected and just disable this for that platform?

davidbanham:
Are we sure it's just win7 though? For all we know it's all the other windows versions too.

davibe:
I did not test Linux and yes, it's quite possible other windows version have the same problem. The guy who requested proxy support is a Windows user in the first place. I would say let's wait a bit because there is a new electron release every ~5 days so maybe next one has the fix.

Or.. i think its possible to work it around some other way because if I disable proxy support, let the app run and then I open dev tools i can do 

    require('remote').require('app').resolveProxy(...) 

and it works fine. Maybe it works when called from another thread .. or being called after <something> has been initialised fixes it. Need to investigate more. But first, again, i would wait for electron.

maxkueng:
> I would say let's wait a bit because there is a new electron release every ~5 days so maybe next one has the fix.

Is there already an issue in the Electron process? I couldn't find it. If nobody reports it's most likely not going to get fixed.

If it's a Windows-ony issue, is there a reason why it can't be enabled for the other platforms?

davibe:
- yes there was an electron issue, its now closed. I don't remember which one was it must be.
- ideed it can be disabled only on failing platforms
- there have been 2 releases of electron since last time i updated this so we should test with newer ones and see if the bug is still there https://github.com/atom/electron/releases

PR #19, opened 2015-06-24
Proxy support

davidbanham:
This is untested as, for some reason, the compiled Electron app produced by ./deploy.sh refuses to run on my laptop. No idea why! That is a problem for another day.

There's the potential for a race condition here if the app attempts to communicate before the resolveProxy calls return. They should complete pretty quickly, however, and any errors will hopefully be ironed out by the retry logic. If there's a more intelligent place to move the calls to to ensure they execute before any network traffic is attempted, by all means do so.

Ref #3

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -59,11 +59,11 @@ app.on 'ready', ->
 
         # Format of proxyURL is either "DIRECT" or "PROXY 127.0.0.1:8888"
 
-        app.resolveProxy 'https://google.com', (proxyURL) ->
+        app.resolveProxy 'https://plus.google.com', (proxyURL) ->
             return if proxyURL is 'DIRECT'
             process.env.HTTPS_PROXY ?= proxyURL.split(' ')[1]
 
-        app.resolveProxy 'http://google.com', (proxyURL) ->
+        app.resolveProxy 'http://plus.google.com', (proxyURL) ->
             return if proxyURL is 'DIRECT'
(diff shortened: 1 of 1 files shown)


algesten:
hey why don't you add these things yourself. sent invitation. the more the merrier. :D

davidbanham:
I see you have foiled my sneaky plan to get you to test this for me. Well played!

Worked around the segfault. Logged the issue with Electron here: https://github.com/atom/electron/issues/2050

All of this appears to be functioning correctly. The only way I have to test this is by using Charles to proxy my connection locally. This technique has served me well in the past, however there is some serious Wild West stuff going on with proxies in educational and corporate environments. Alls I'm saying is we shouldn't be shocked to see edge cases from users.

davidbanham:
I'll merge this in as soon as I've got a :+1: from an existing contributor that we're happy with the code style, etc  (4 space indentation? You monsters!).

algesten:
how worried are we about that race condition? I mean if we have the slightest concern, it's pretty darn easy to ensure there's no network activity too early.

the entire hangupsjs starting point is this row:

        # first connect
        reconnect().then ->

but if we're not worried then lgtm

+1

davidbanham:
Ah! Yes. If it's that straightforward then yes, absolutely makes sense to delay starting the network activity until the proxy stuff is done. I was basically being lazy and not investigating how the network stuff was kicked off.

I'll make the changes.

algesten:
Sweet! +1

PR #23, opened 2015-06-27
Window Icon

maxkueng:
This copies the icons in to the app directory (in gulp) and uses them as the window icon. It should work on all platforms but I have only tested it on Linux.

I had to move `doFirstConnection` further down as it was called before it was defined and resulted in a crash.

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -2,6 +2,7 @@ gulp   = require 'gulp'
 coffee = require 'gulp-coffee'
 less   = require 'gulp-less'
 rimraf     = require 'rimraf'
+path       = require 'path'
 fs         = require 'fs'
 gutil      = require 'gulp-util'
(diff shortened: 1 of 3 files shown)


algesten:
looking good. i don't quite get the icon jiggling. but hey. if it works it works :)

maxkueng:
By "icon jiggling" you mean the `@2x` stuff? This is documented in the [Navtie Image](https://github.com/atom/electron/blob/master/docs/api/native-image.md#high-resolution-image) docs. It's the DPI multiplier for high-resolution images. Retina and such. By naming the file `icon@2x.png` Electron will automatically use this image instead of `icon.png` on displays with twice the DPI.

I'm sure the copying could have been done more elegantly in the Gulpfile. This is the first time I did anything like that in Coffeescript.

PR #25, opened 2015-06-30
Extend .editorconfig

maxkueng:
I've added encoding and line endling settings to .editorconfig. Also added indentation size for package.json and .travis.yml. Despite the other coffeescript code using 4 spaces I saw that gulpfile.coffee uses 2 spaces so I've also added that.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ indent_style = space
 [*.coffee]
 indent_size = 4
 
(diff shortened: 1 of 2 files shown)


algesten:
i invited you so you can fix this yourself. reindenting the gulpfile would be my preference.

maxkueng:
Okay, I've fixed the indentation in the gulpfile.

PR #28, opened 2015-07-02
Use Electron v0.27.3 because of a copy paste issue in v0.28

### LINKED ISSUE #27: Double clipboard content on paste
When i paste my clipboard text into the message textfield the text gets pasted twice.

maxkueng:
This is supposed to fix #27 temporarily.


algesten:
i just upgraded to 0.30.1

Issue #43, opened 2015-08-10
Minimize to tray/start minimized


Georgeto:
It would be awesome, if it was possible to minimize yakyak to tray, and hide it from taskbar while being minimized. In combination with that feature, a 'start minimized' option would be handy.

algesten:
None of the more active devs have currently windows as their main platform. This project could really do with someone helping out on that side.

Georgeto:
I've tried to implement it by myself, but it didn't work out as expected. The problem is hiding the window from taskbar. I've done it through calling mainWindow.hide() on minimize event, but immediately after the window is hidden, it pops up again, but without content.

*main.coffee*
```coffee
...
Tray      = require 'tray'
...
tray = null
...
app.on 'ready', ->
    ...
    mainWindow.on 'minimize', ->
        tray = new Tray path.join __dirname, 'icons', 'icon.png'
        mainWindow.hide()
```

![blackyakyak](https://cloud.githubusercontent.com/assets/9250103/9179779/1278ddd6-3f9f-11e5-8071-72afcd01be03.PNG)

maxkueng:
Linux has trays too (sometimes) and as far as I know (but I'm not sure) the OS X Dock basically acts as a tray because on OS X apps typically don't really exit when you close the window. But I may be talking nonsense.

Yes, would be nice if more Windows people came to help out.

maxkueng:
@Georgeto: I'm not sure if the `minimize` event is the right place to hide the window. I think the regular minimize (to taskbar) and minimize to tray should be treated as different things.

Typically, apps that support tray icons do it as follows:

 - The minimize button in the window  title bar minimizes the window to the taskbar just like any other window.
 - Clicking the tray icon toggles window visibility and when hidden it also disappears from the task bar. When shown it also reappears in the task bar.
 - There is an setting/option whether to just minimize to tray when closing the window (through "×" button in the window title bar), or whether to actually exit the app. It's often labeled as "Keep running in the system tray when closed". Minimize to tray is usually the default.
 - Right-clicking the tray icon usually opens a context menu that offers additional functionality including showing the window if it's hidden, hiding it if it's visible, and actually exiting the app.

Georgeto:
@maxkueng: It was just a quick test, that I have done to play around with the tray object. The behavior  you described is exactly what I requested and its a shame, that I did not described it in such detail.

maxkueng:
[...]

PR #47, opened 2015-08-12
YakYak to the tray

Georgeto:
Like discussed in #43 I've added a basic implementation, which offers the following features.
>
* The minimize button in the window title bar minimizes the window to the taskbar just like any other window.
* Clicking the tray icon toggles window visibility and when hidden it also disappears from the task bar. When shown it also reappears in the task bar.
* There is an setting/option whether to just minimize to tray when closing the window (through "×" button in the window title bar), or whether to actually exit the app. It's often labeled as "Keep running in the system tray when closed". Minimize to tray is usually the default.
* Right-clicking the tray icon usually opens a context menu that offers additional functionality including showing the window if it's hidden, hiding it if it's visible, and actually exiting the app.

(Quoted from @maxkueng)

I have only tested my changes on windows and I'm completely new to coffeescript, nodejs and electron, please take this into consideration when looking over my code.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -6,6 +6,8 @@ fs        = require 'fs'
 path      = require 'path'
 tmp       = require 'tmp'
 clipboard = require 'clipboard'
+Tray      = require 'tray'
+Menu      = require 'menu'
 
 tmp.setGracefulCleanup()
 
@@ -17,6 +19,7 @@ paths =
     rtokenpath:  path.normalize path.join app.getPath('userData'), 'refreshtoken.txt'
     cookiespath: path.normalize path.join app.getPath('userData'), 'cookies.json'
     chromecookie: path.normalize path.join app.getPath('userData'), 'Cookies'
+    configpath: path.normalize path.join app.getPath('userData'), 'config.json'
 
 client = new Client
     rtokenpath:  paths.rtokenpath
@@ -36,12 +39,31 @@ logout = ->
         cwd: process.cwd
         env: process.env
         stdio: 'inherit'
-      app.quit()
+      quit()
     return promise # like it matters
 
 seqreq = require './seqreq'
 
+saveConfig = ->
+    nconf.save (err) ->
+        console.log 'error while writing config.json', err if err
+
+# Load configuration
+nconf = require 'nconf'
+nconf.file { file: paths.configpath }
+nconf.defaults {
+    'startminimized': false
+    'minimizetotray': false
+}
+
 mainWindow = null
+tray = null
+
+# No more minimizing to tray, just close it
+readyToClose = false
+quit = ->
+    readyToClose = true
+    app.quit()
 
 # Quit when all windows are closed.
 app.on 'window-all-closed', ->
@@ -49,6 +71,9 @@ app.on 'window-all-closed', ->
 
 loadAppWindow = ->
     mainWindow.loadUrl 'file://' + __dirname + '/ui/index.html'
+
+toggleWindowVisible = ->
+    if mainWindow.isVisible() then mainWindow.hide() else mainWindow.show()
 
 # helper wait promise
 wait = (t) -> Q.Promise (rs) -> setTimeout rs, t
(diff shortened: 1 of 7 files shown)


maxkueng:
Amazing! I don't know CoffeeScript well enough to judge the quality of your code (but it looks good to me). But I've pulled your branch and the tray icon works perfectly also on Linux. It does all the things described. Awesome job!

I've noticed one thing:

Now that Yakyak can keep running in the background there two ways to close the app: Just close the window, and actually quit. From the "Yakyak" window menu only "Quit" _(Ctrl+Q)_ is available. I think it would make sense if there was another entry to just close the window called "Close" _(Ctrl+W)_ - maybe put it above the "Quit" entry.

Can somebody try this out on OS X and see whether it works or causes any conflicts with the Mac UX, and look at the code?

src/main.coffee
```diff
@@ -73,8 +98,18 @@ app.on 'ready', ->
...
+        show: !nconf.get 'startminimized'
     }
 
+    # Create tray
+    tray = new Tray path.join __dirname, 'icons', 'icon.png'
+    contextMenu = Menu.buildFromTemplate [
+        { label: 'Hide/show', click: toggleWindowVisible }
+        { label: 'Quit', click: quit}
+    ]
+    tray.setToolTip 'YakYak - Hangouts client'
```
maxkueng: Yakyak is written with the second "y" lower cased. "Yakyak" not "YakYak".

algesten:
Good work! However I couple of issues before we can land it.

1. "Command" is indeed called that on Mac OS X. "Control" is a windows term. This needs to be set correctly depending on platform. ec15a60bf
2. The introduction of a server side config file means we've split the location for such things and now have configs in two places. 46b30620 – the other place is in [src/ui/model/viewstate.coffee](https://github.com/yakyak/yakyak/blob/master/src/ui/models/viewstate.coffee#L74-L87) where we keep things like window position, size etc in `localStorage`. This needs consolidating so that config is only stored in one place. I'm not against moving it to `config.json` on the server, but then all needs to move there. I suspect the best would be to have a ipc call that stores all such values in one go (rather than separate per config option).
3. I didn't actually run the code, but "minimize to tray" has no Mac OS X equivalent. So I suspect the best option would just be to hide that config option from the user. Similar to point 1, detect platform, hide if not windows/linux.
4. Regarding YakYak – we certainly started with a lowercase y, @maxkueng is right. However it seems we're going towards an uppercase since this would be better for english native speakers.

maxkueng:
1. Electron has a [modifier](https://github.com/atom/electron/blob/master/docs/api/accelerator.md#available-modifiers) called `CommandOrControl` thatS X and `Control` on Windows and Linux. We might be able to use that without making completely separate menus for each platform.

 2. I think it makes sense to have separate places for actual configuration and view state. Things like the current window size and position and currently selected conversation aren't actual settings. They're more like the current state of the UI which can be persisted. But whether the app should always start minimized to the tray is an actual runtime configuration setting so it belongs somewhere else. So in my opinion the view state shouldn't be moved in to the "server side" config file.  
Instead of using `ipc` there's also the option to use `global`. If you do `global.myObj = myObj` in the browser context you can access the object in the renderer context using `var myObj = require('remote').getGlobal('myObj')`

 3. .

 4. I think the name can be changed from "Yakyak" to "YakYak" once the decision has been made to change it. But for now using lower-case should be fine.

algesten:
1. ok. use that :D
2. ok. we could make distinction between view state and preferences (although i suspect it's a blurry line at best), we do have "Show conversation thumbnails" as a view setting. this one is stored in localStorage.
3. .
4. The decision has been made. YakYak it has been for quite some time. Other ways of spelling it is just @davibe being confused ;)

maxkueng:
Okay, so as I see it the only thing that needs to be done to merge this is:

 - Only start the tray on Windows and Linux (i.e. ` if require('os').platform() !== 'darwin'`)

All the other things are not directly linked to this issue but are general problems and can be done later:

 - Unify the menu as much as possible and use the `CommandOrControl` modifier for accelerators. Separate menu templates for OSX and non-OSX have existed before and were not introduced in this branch.
 - Use consistent naming.
 - Distinguish between preferences and persistent view state. I also agree that "Show conversation thumbnails" is a preference and not a view state.

davibe:
any update on this feature ?

maxkueng:
Seems like @Georgeto has lost interest. I'll merge this in to a separate branch and fix the `if require('os').platform() !== 'darwin'` issue. Then merge it in to master from there.

Regarding the other issues the have been discussed here, they seem to be general problems and not really related to this issue. So they can be fixed later. I'll put them in to separate issues.

PR #51, opened 2015-09-01
Add handler for the 'appfocus' event

### LINKED ISSUE #49: Clicking notification doesn't focus window
On Windows when a notification popup comes up clicking the notification doesn't bring forward the app window.

There seems to be code written to implement that feature [here](https://github.com/yakyak/yakyak/blob/22312d8fa55b26197cb475c5ed08b3e293d943df/src/ui/views/notifications.coffee#L63) however the `res` is looking for "Activate" and in my experimentation on Windows it will be "Activated".

But *even if* I fix that line `action 'appfocus'` doesn't actually *do* anything. It seems to do an IPC call to (I assume) electron but my understanding ends there.

Is this an electron issue? Or is yakyak not calling the right IPC function [here](https://github.com/yakyak/yakyak/blob/22312d8fa55b26197cb475c5ed08b3e293d943df/src/ui/dispatcher.coffee#L99)? Does this work on mac?

xori:
Fixes #49

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -208,7 +208,9 @@ app.on 'ready', ->
         client.setfocus conv_id
     , false, (ev, conv_id) -> conv_id
 
-    ipc.on 'appfocus', -> app.focus()
+    ipc.on 'appfocus', ->
+      app.focus()
+      mainWindow.focus()
 
     # no retries, dedupe on conv_id
     ipc.on 'settyping', seqreq (ev, conv_id, v) ->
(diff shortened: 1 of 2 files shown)


xori:
I kept the existing `app.focus()` call because maybe that actually works for Linux/Mac but on Windows it does nothing.

algesten:
could you merge the commits to one? otherwise lgtm

xori:
Rebased.

algesten:
sweet. thanks!

PR #52, opened 2015-09-01
If an avatar fails to load, switch to a local fallback.

### LINKED ISSUE #41: Private avatar images broken
If a user's avatar image is private it returns a 403 response, which displays as a broken image in the interface.

Here's the image, source, and google's response all crammed into one screenshot:

https://dl.dropboxusercontent.com/spa/wn0oryty1do7jho/u_df5zyp.png

xori:
fixes #41 

not sure if this is the best way to do it, or if I should check in the lines before, but it fixed it for me.


algesten:
lgtm

PR #53, opened 2015-09-04
Hide windows instead of close for OS X

alexbragdon:
On OS X: Closing the main windows no longer closes the app and simply hides the main windows.
 
Clicking on the dock icon now unhides the window if it's hidden. Closing of the app is now only accomplished through the menu bar or right clicking on the dock icon.

There is no issue for this and I'm not sure if anyone else thinks this is how window management should work (though I suspect from the comment on the old line 48 that someone does) but here it is if you guys want it. OS X only as it's what I use, though I could test elsewhere if needed. Else, please ignore me and have a nice day!

-Alex

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -45,7 +45,15 @@ mainWindow = null
 
 # Quit when all windows are closed.
 app.on 'window-all-closed', ->
-    app.quit() # if (process.platform != 'darwin')
+    app.quit() if (process.platform != 'darwin')
+
+# For OSX show window main window if we've hidden it.
+app.on 'activate-with-no-open-windows', ->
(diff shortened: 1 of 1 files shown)


maxkueng:
AFAIK that's the standard behavior for most apps on OS X. Closing the window doesn't quit. To quit you have to press Command+Q or use the menu.

alexbragdon:
Yup it is the standard behavior. Which is why I found it a little annoying when it didn't work.

algesten:
lgtm

maxkueng:
Oh, this is a PR. I didn't look properly and I thought it was already doing that :blush:

PR #54, opened 2015-09-05
Linux: Added Ctrl-J/K for conversations, fixed keybindings

mkzero:


mkzero:
@Tux64 I'm not very familiar with the Windows world, but isn't the Alt+F4 handled by Windows window manager?

maxkueng:
@Tux64, @mkzero: I thought so, too. `Alt+F4` should be handled by Windows.

What's the purpose of the Makefile? There are already Gulp tasks for everything it does. Wouldn't it make more sense to add those things to the `package.json` as "run scripts". Make, for example, can't access the local Gulp binary so it has to use the globally installed package, use use `node_modules/gulp/bin/gulp.js`. npm on the other hand will automatically use the local binaries.

mkzero:
Sorry, forgot to create a branch. The makefile was for myself... m( That commit shouldn't have been in the PR. You can cherry-pick 339af7a and e1a65ac from the PR.

PR #60, opened 2015-09-21
workaround fixing double clipboard content on paste

franc90:
Solves #27 - tested only on Manjaro Linux. I believe it will work on other Arch-based distributions.

The error occurs because two onpaste events are fired. 

I am neither JS nor CoffeeScript developer and I know this is a hideous solution to the problem but with my limited knowledge it must be sufficient. I hope that after discovering the issue someone will fix it in a better way, and this will be used as a temporary solution by those who suffer from double content paste.


maxkueng:
It works. Thanks, mate!

maxkueng:
fixes #27

Issue #24, opened 2015-06-29
icon on windows needs improving


algesten:
![tmp-52438kggn14bu58l3](https://cloud.githubusercontent.com/assets/227204/8416305/185d9ec8-1ea5-11e5-9d30-46e4f7883e99.png)

maxkueng:
Maybe the standard icon needs to be larger. Can you try to play around with the sizes? What if the `icon.png` was 64px? Does it still look good in in the window title? I don't have a Windows computer and my Linux desktop doesn't have window titles. I only have a small taskbar-like thing.

![yakyak-taskbar](https://cloud.githubusercontent.com/assets/168709/8427000/589c69f0-1f14-11e5-93fd-80766de923b5.png)

algesten:
i don't use windows myself. but yes. i think it should be one of the larger sizes.

Gibbz:
yeah it looks horrific on ubuntu also. Looks very low resolution

AdrianoCahete:
And if you have a dark wallpaper, its seems to quite off/dark...

![yakyak_win10.PNG](https://files.zenhub.io/566582ddb51138092d360b6e)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

PR #99, opened 2015-12-18
WIP: Enable simple tray icon state changes

### LINKED ISSUE #98: Menu bar icon for OS X
I'd like to have an Adium-style menu bar icon for YakYak as an alternative notification indicator to dock icons or the OS X Notification Center. Just having a couple of visually distinct states (unread vs. no unread) would be perfectly fine to start (unread message count would be a bonus).

ironcladlou:
### :construction: WIP :construction: 
Add a very simple tray icon state change: when there are unread messages, use a distinct tray icon to provide a visual cue.

Fixes #98 

TODO:
- [ ] Better icon (probably depends on #24)
- [ ] Test `deploy.sh` changes on other platforms
- [ ] Get an explanation for the darwin tray exclusion in https://github.com/yakyak/yakyak/commit/ecf7e02df32110b51c4a76a2f8d71a2aba3082ba (cc @maxkueng)

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,4 +1,31 @@
 #!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT=$(dirname "${BASH_SOURCE}")
+
+function yakyak::package() {
+  local arch="$1"
+  local src="electron"
+  local dest="yakyak"
+
+  if [[ ! -f "$src" ]]; then
+    echo "Skipping packaging for $arch (source dist not found)"
+    return
+  fi
+
+  if [[ "$arch" =~ win32 ]]; then
+    src="${src}.exe"
+    dest="${dest}.exe"
+  fi
+
+  pushd $arch >/dev/null
+    mv $src $dest
+    cp -R ../../app resources/app
+  popd >/dev/null
+  zip -r yakyak-$arch.zip $arch
+}
 
 for dep in curl unzip sed; do
   echo "checking dependency... $dep"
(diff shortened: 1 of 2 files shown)


maxkueng:
@ironcladlou: The tray feature is disabled on OS X because there is no such thing as "minimize to tray" on OS X, according to @algesten's [comment in PR #47](https://github.com/yakyak/yakyak/pull/47#issuecomment-132073116) (point 3).  
I don't have OS X so I have no idea. But I believe it isn't a thing because on OS X apps don't really exit when you close the window. Instead, they keep running until you `Cmd+Q` or select "Quit" from the menu. That might be the Mac version of "minimize to tray" and I think most desktop software does that.

algesten:
looks good. @maxkueng is right. osx hasn't got the exact equivalent of a tray. there is the system icon area, where a lot of programs (erroneously) belive they belong. I'm sure some people think they want yakyak there.

davibe:
Thank you for your PR, sorry for merging so late !

ironcladlou:
@maxkueng
>@ironcladlou: The tray feature is disabled on OS X because there is no such thing as "minimize to tray" on OS X, according to @algesten's comment in PR #47 (point 3).
>I don't have OS X so I have no idea. But I believe it isn't a thing because on OS X apps don't really exit when you close the window. Instead, they keep running until you Cmd+Q or select "Quit" from the menu. That might be the Mac version of "minimize to tray" and I think most desktop software does that.

Clicking "Hide/show" from the tray icon context menu works as expected on OSX. The main window becomes hidden/unhidden.

@algesten
>looks good. @maxkueng is right. osx hasn't got the exact equivalent of a tray. there is the system icon area, where a lot of programs (erroneously) belive they belong. I'm sure some people think they want yakyak there.

[Adium](https://adium.im) uses a menu icon, as does [MailPlane](http://mailplaneapp.com). There's a ton of precedence out there for OS X apps using the menu area for notifications. OS X has a notification subsystem, but it's not great for this class of application.

ironcladlou:
Thanks for the merge... I didn't expect it to happen without some more discussion, otherwise I would have squashed my commits. Sorry about that.

davibe:
I think we are missing the 2 images. Can you also push those ?

ironcladlou:
@davibe 

Added the missing files in https://github.com/yakyak/yakyak/pull/112. I took care to very clearly mark this PR as a WIP; I don't think it should have been merged until it was reviewed and I had a chance to squash and do more testing. I believe the deploy script also has some breakages from this that I didn't catch until doing a linux build.

davibe:
I already reverted the build script as it was not working. I know I merged too early. It's not a big problem though. The main problem is the lack of PRs and limited time of the original developers. Anything that makes things move is good anyway.

PR #100, opened 2015-12-29
Replace icon to hangout logo

tomcsi:


kethinov:
There may be some trademark issues associated with this.

maxkueng:
I'm also pretty convinced we can't just use Google's logo without their consent.  
But what we can do is make a logo that looks more like the logo of a chat app, . There's already some work going on but it seems it has stalled.

PR #105, opened 2016-01-07
Add extra hints in README.md for new developers

mhogerheijde:
Not knowing gulp and electron beforehand, could make it a bit difficult
to start out. It took me quite some fiddling and looking at misleading
error messages to figure out I should have installed `gulp` _before_
running `npm install` (whereas I assumed npm would install it for me).

The electron binary, also, isn't directly available as a command since
npm does not put it on the $PATH for you. So I added a hint to where the
binary is located.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ It is written in coffeescript (nodejs) based on
 ### Setup
 
 ```bash
+$ sudo npm install -g gulp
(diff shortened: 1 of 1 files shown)


davibe:
ty

mhogerheijde:
np

Issue #34, opened 2015-07-12
Characters Not Supported


WhiteHatHacker1:
Many non-standard characters including emoji are not supported and show up as blank rectangles.

algesten:
emojis are a special case since we may want to use a specific font for exactly those unicodes. beyond that we use standard fonts which may then lack characters. what platform are you on and can you give an example character that is missing.

WhiteHatHacker1:
Thank you for responding so quickly!
I'm on an Arch Linux variant. The characters I've noticed missing are some Greek ones such as the letter epsilon (ɛ) or alpha (ɑ) and other characters of the IPA Extensions and Spacing Modifier Letters Unicode blocks ([see here]( https://en.wikipedia.org/wiki/Phonetic_symbols_in_Unicode#From_Unicode_blocks_to_scripts)).

brendan-benting:
I'm using Ubuntu (Gnome) 15.04 64 bit, Chinese Traditional characters are being rendered as blank rectangles as well.

algesten:
This will be down to the font installed on those systems not having the glyphs for those characters. Our font string is currently:

```
    font-family: "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif;
```

Are there better options for these systems. Or is the solution to ship a (free) font that has the glyphs – and in that case, which font?

brendan-benting:
Google noto has these glyphs, open source and attractive to boot.

algesten:
I tried notosans in a separate branch https://github.com/yakyak/yakyak/tree/notosans – however I think it looks ridiculous.

<img width="960" alt="screen shot 2015-07-26 at 10 07 29" src="https://cloud.githubusercontent.com/assets/227204/8893032/45b77a18-337e-11e5-9e11-30e88bb5a7aa.png">

brendan-benting:
Would it be possible for adding more fonts to the font-family string:

宋体, SimSun -- OSX and Windows at least will have it, and those using linux have already installed extra chinese fonts themselves.

algesten:
does the font fall back if the current lacks glyphs for the character. that would be good in that case.

jorgeavaldez:
Wouldn't this work? http://emojione.com/

davibe:
can someone try to set notosans or any other richer font as a fallback for the current one and see if that works with special characters ? In that case a pull request would be trivial.

DASPRiD:
[...]

PR #108, opened 2016-01-09
Add emoticon-selector and NotoColorEmoji font

### LINKED ISSUE #42: Smileys/Emoticons not displaying
The smileys and emoticons are not displaying in either the chat windows or in the chat headings. It just shows a small rectangle.

Also theres no button on the interface to add them either :)

mhogerheijde:
This Pull-Request introduces a selector for the first code-block of emoji / emoticons (1F600 trough 1F64F) and also introduces [NotoColorEmoji](https://www.google.com/get/noto/#emoji-qaae-color) font, an open font from Google.

The new font is used in the selector, message input field and for the text of send/received messages.

This should resolve #42, and a bit of #34.

I did not (yet) add all the code blocks, since I'd like to have some feedback on this first. If it fits in your view/roadmap, whether my proposed emoticon-selector works good (enough), etc.. After that, I'd gladly add the rest of the code-pages to the selector.

"known issues":

1. Emoticons are added in the textfield /after/ the cursor. If there is a selection, it will not overwrite the seletion like what would happen when you input a character using the keyboard.
2. The selector does not automatically close, you'll have to push the button again.


src/ui/css/fontello/css/fontello.css
```diff
@@ -74,5 +74,6 @@
 .icon-bell:before { content: '\e811'; } /* '' */
 .icon-bell-off-empty:before { content: '\e812'; } /* '' */
 .icon-attach:before { content: '\e813'; } /* '' */
+.icon-emoji:before { content: '\1f603'; font-family: "Helvetica Neue"} /* '' */
```
mhogerheijde: I didn't think it would fit to use Noto here, since the "attach" icon is monochromatic also.

mhogerheijde:
I just noticed an issue with spaces not being rendered.. It seems that the space character in the Noto font is zero-with?! The NotoColorEmoji font should not provide "normal" letters, so it should only be used to render the emotiocon range,.. right? Or am I misunderstanding the fall-back mechanism of the `font-family` property?

I would expect that there is no space character in the NotoColorEmoji font, or – at least – it to be non-zero width.

davibe:
I tried your branch. You're definitely on the right track. I have some graphical comments

screenshot 2016-01-11 11 13 46

I zoomed a portion of yakyak. Here are my observation
- There is a double border. I think if you remove the rounded one is going to be ok. 
- There is also a top margin of some sort that covers part of the conversation clipping away thins that I could be seen.
- I think in the drawer the emoticons could be bigger with less margin around them
- the image inside the button revealing the drawer is not well centered vertically 

Other than that its looking good and If you can figure out spaces and some other emoticon group it would be a great contribution to yakyak.

mhogerheijde:
Thanks for the feedback. I'll try to fix all the margin/border issues etc.

Interesting thing: these emoticons look different at my machine? Mine are like these Noto Emoji

I'll post a screenshot after I've made some fixes.

mhogerheijde:
I've updated the styling using your suggestions and I fixed the spaces. Only thing left is the other emoticon-groups.

davibe:
<3

davibe:
news on the other groups ? :)

mhogerheijde:
Sorry, had a busy week and – moreover – a busy weekend :) 

Current status: meddling with closures to get the correct button to open the correct pane of emoji. When that works it should be little effort to create multiple panes of Emoticons. 

Things on the to-do list:
1. Create opinion about the grouping of emoji. The unicode.org grouping seems a bit on the awkward side. I've defined the groups in a mapping, so changing the order/grouping should be just changing the mapping without any other work.
2. Design of the drawer + tabs. I'm not a designer, so I'll try my best to get it to look as good as possible, but there probably will be some comments/feedback.

mhogerheijde:
Okay, so: if we look at unicode.org and the codepoints they define under "Emoji & Pictographs" (see http://www.unicode.org/charts/ ), there are 5 groups of code-points in unicode (not counting dingbats)

1. Emoticons [0x1F600..0x1F64F]
 - Fully present in Noto
2. Miscellaneous Symbols [0x2600..0x26FF]
 - Fully present, but contains a lot of awkward glyphs we might want to leave out
3. Miscellaneous Symbols And Pictographs [0x1F300..0x1F5FF]
 - Mostly present, some missing and a few awkward glyphs
4. Supplemental Symbols and Pictographs [0x1F900..0x1F9FF]
 - Only a few present, a few that I would categorise under 1) and a few animal emoji
5. Transport and Map Symbols [0x1F680..0x1F6FF]
 - Mostly present and a few awkward glyphs.

Maybe I shouldn't just take the ranges as they are defined by unicode.org and make our own categories. (luckily I made that quite simple to do).

I'll also take out / blacklist codepoints that don't have a glyph in the font. I don't know if I could automate the extraction of all supported codepoints in a font, would be nice to know beforehand if a codepoint is represented by a font or not. (to prevent the empty box glyph)

davibe:
http://emojipedia.org

mhogerheijde:
I think these categories are quite nice! They differ from both _Hangouts_ itself and _WhatsApp_, but they differ between themselves already, so I don't think it is necessary to have the same categories.

I'll try and contact them to see if they have a machine-readable version of their categorisation, otherwise I'll write a quick-n-dirty scrape script to get the categorisation from HTML or something.

In the meanwhile; I currently have this:

selector

This works, technically, but it sill looks ugly :) I'll try to add better padding and spacing, but it's probably best if someone with a better understanding of design & UX takes a look at this.

The box-glyph will disappear when I get/use the categorisation from [emojipedia.org](http://emojipedia.org)

davibe:
the first line are the last used ones ?

mhogerheijde:
Right, still not mergeable, but I have scraped the categories from [emojipedia.org](http://emojipedia.org); their categories are far better.

A new issue arose, where there are multi-codepoint emoji (with a glue-codepoint), that should render as 1 glyph, but currently don't render as one for the <span> elements, but _do_ render as one in the <input> element.... which is strange.

I've created a python-script to scrape [emojipedia.org](http://emojipedia.org) and to create a big mapping in `JSON`. Maybe it should be converted to coffeescript and made part of the build process? Or maybe a gulp target to easily fetch changes from [emojipedia.org](http://emojipedia.org).

Anyway, I'll update when I have some new progress.

mhogerheijde:
> the first line are the last used ones ?

No those would be the categories to switch the code-pages.

davibe:
I would not put the scraping as a part of the build script. Maybe we should commit the script with the result and just update it manually from time to time for now

mhogerheijde:
By "made part of the build process" I actually meant: exposing it as a gulp target. So you can do something like `$ gulp update-emoji-definition` and it just runs the script. It should, indeed, only be ran manually.

mhogerheijde:
I guess I'm done! :) At least I think it is acceptable for a first iteration?

emoticonselector

I've included the python-script in this PR, but it needs some dependencies, so for a non-python developer it is not clear how to set it up. Maybe it should be rewritten into coffeescript to keep the development tools clean & lean.

davibe:
cool :) will try it in the next days

davibe:
I just noticed there is a native thing on OS X that allows you to input emojis. It opens as a separate window like this. 
screenshot 2016-02-10 18 18 14

davibe:
which os do you use ?

mhogerheijde:
I'm on Linux, Ubuntu, no built-in keyboard like that, but a vast support for unicode, so it probably is doable to get a emoji-keyboard-thingy. But still: I think having it inside YakYak is nice :smile:

davibe:
indeed its merged :)

mhogerheijde:
I noticed :)

Maybe there is a way to detect if the OS is capable of showing the selector? And just open that one when clicking the button, otherwise use the drawer as a fall-back. I'll try to find out, I'll probably be able to get my hands on a Mac and see.

davibe:
I am sure its possible. But using a separate window or not should probably be up to the user so i would leave it like it is right now.

PR #117, opened 2016-01-22
allow URLs to break in messages

willwhitney:
Previously URLs that contained no breaking characters would not wrap.

Before:
screenshot 2016-01-22 11 58 37

After:
screenshot 2016-01-22 11 34 12


davibe:
super, thank you

PR #128, opened 2016-02-07
[Mac] - Add ability to hide Dock Icon, Minimize when started

Derimagia:
It looks like there are a few settings that don't exist on Mac but were implemented for Windows.

Namely:
1) Minimize when started. I was able to enable this by just enabling the checkbox.
2) Ability to minimize to tray. On Mac, this would be hiding the dock icon. I implemented this change, and added the setting to the context menu. This setting should probably be moved to a settings in the app, but right now there isn't a settings page. I couldn't put it in the same place as the other settings because when you hide a dock icon, it also hides the menu bar for the app.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -83,10 +83,21 @@ loadAppWindow = ->
 toggleWindowVisible = ->
     if mainWindow.isVisible() then mainWindow.hide() else mainWindow.show()
 
+toggleDockIcon = ->
+    if nconf.get 'minimizetotray'
+      app.dock.show()
+      nconf.set 'minimizetotray', false
+    else
+      app.dock.hide()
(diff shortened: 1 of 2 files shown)


kecinzer:
Hey! Can you please make also tray icon hiddable?

Derimagia:
Main issue I see with making the tray icon hidable is that we need to make sure they don't make the dock icon hideable and the tray hideable or else they won't have a way to get back in the app. I don't think it's possible on a Mac but it may be possible on a Windows machine since the settings for minimizing to tray is on the Title bar.

This is a little trickier since there isn't a settings page (I had to add the dock icon settings to the tray icon context menu, for example, to prevent this from happening)

PR #131, opened 2016-02-11
tray: fix build context menu

### LINKED ISSUE #130: Unable to connect after update to 1.1.3
Hi there,

So I just update to the latest version and now I'm unable to get a connection or something.
Here is this log:
```
input.coffee:43 Why is this running multiple times?
input.coffee:46 Creating selector for people
input.coffee:46 Creating selector for nature
input.coffee:46 Creating selector for food
input.coffee:46 Creating selector for activity
input.coffee:46 Creating selector for travel-places
input.coffee:46 Creating selector for objects
input.coffee:46 Creating selector for symbols
input.coffee:46 Creating selector for flags
/opt/yakyak/resources/atom.asar/renderer/api/lib/remote.js:109 Uncaught Error: Cannot read property 'position' of undefined
TypeError: Cannot read property 'position' of undefined
    at Function.Menu.buildFromTemplate (/opt/yakyak/resources/atom.asar/browser/api/lib/menu.js:291:15)
    at callFunction (/opt/yakyak/resources/atom.asar/browser/lib/rpc-server.js:148:18)
    at EventEmitter.<anonymous> (/opt/yakyak/resources/atom.asar/browser/lib/rpc-server.js:234:14)
    at emitMany (events.js:108:13)
    at EventEmitter.emit (events.js:182:7)
    at EventEmitter.<anonymous> (/opt/yakyak/resources/atom.asar/browser/api/lib/web-contents.js:97:23)
    at emitTwo (events.js:87:13)
    at EventEmitter.emit (events.js:172:7)
input.coffee:43 Why is this running multiple times?
```

I just keeps looping with the `input.coffee:43 Why is this running multiple times?` as start.

I'm running ubuntu 15.10 with GNOME Shell 3.16.4 and Node 4.3.0

I hope you can help me fix this issue.

Cheers,
Warnar

Georgeto:
I don't see why users not using MacOS should not have a quit option in the tray context menu.

Fixes #130

### DIFF
--- a/src/ui/views/trayicon.coffee
+++ b/src/ui/views/trayicon.coffee
@@ -16,7 +16,7 @@ create = () ->
     tray.setToolTip 'YakYak - Hangouts client'
     # Emitted when the tray icon is clicked
     tray.on 'clicked', -> action 'showwindow'
-    
+
 destroy = ->
     tray.destroy() if tray
     tray = null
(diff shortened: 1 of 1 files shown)


src/ui/views/trayicon.coffee
```diff
@@ -34,20 +34,23 @@ update = (unreadCount, viewstate) ->
           checked: viewstate.startminimizedtotray
           click: -> action 'togglestartminimizedtotray'
         }
-        { 
+        {
           label: 'Hide Dock icon'
           type: 'checkbox'
           checked: viewstate.hidedockicon
           click: -> action 'togglehidedockicon'
-        } if require('os').platform() == 'darwin'
```
davibe: this if was intended to be referred to the Hide Dock Icon which does not make sense on windows

davibe: I think the problems is that the inlined if leaves a `null` when it's false and electron does not like that

Georgeto: Yeah that is the problem, see my screenshot in #130.

PR #134, opened 2016-02-12
Avoid scrollToBottom when messages is undefined

boekkooi:
This happens to be on Gnome when starting the application.

aka this avoids the error:
```
Uncaught TypeError: Cannot read property 'scrollToBottom' of undefined
```


davibe:
thank you

PR #154, opened 2016-03-16
Made close to tray optional for every system

### LINKED ISSUE #135: Unable to close to tray
Hi there,

So in version 1.1.2 I was able ti hit the X (close) button in the window and it would close to the tray icon this function seems to be gone in 1.1.3. 
Could you tell me where to look so i can reintroduce this?

Cheers,
Warnar

boekkooi:
Hi there,

So I finally had some time to fix #135 

By default close to tray is off but can be set to on using the tray menu.

When turned on it will use ['beforeunload'](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#event-close) to avoid closing the window expect when quit triggers the custom window event 'quit'.

Let me know what you think.

Cheers,
Warnar

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -46,22 +46,20 @@ seqreq = require './seqreq'
 mainWindow = null
 
 # No more minimizing to tray, just close it
-readyToClose = false
+global.forceClose = false;
 quit = ->
-    readyToClose = true
+    global.forceClose = true;
     app.quit()
+    return
 
-# Quit when all windows are closed.
-app.on 'window-all-closed', ->
-    app.quit() if (process.platform != 'darwin')
+app.on 'before-quit', ->
+    global.forceClose = true;
+    return
 
 # For OSX show window main window if we've hidden it.
-app.on 'activate-with-no-open-windows', ->
-    mainWindow.show() if (process.platform == 'darwin')
-
-# If we're actually trying to close the app set it to force close
-app.on 'before-quit', ->
-    mainWindow?.forceClose = true
+# https://github.com/electron/electron/blob/master/docs/api/app.md#event-activate-os-x
+app.on 'activate', ->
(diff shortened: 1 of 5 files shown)


algesten:
can you resolve the conflicts please?

boekkooi:
@algesten Rebased

algesten:
There is some problem with this commit.

On OS X if I swith off "close to tray" and press the close button, the logs spews out a bunch of

```
2016-04-08 21:48:09 INFO poller stopped Cannot read property 'webContents' of null
connect_failed
2016-04-08 21:48:09 INFO poller stopped Cannot read property 'webContents' of null
connect_failed
2016-04-08 21:48:09 INFO poller stopped Cannot read property 'webContents' of null
connect_failed
```

and if i click the tray icon again, i get a 

screen shot 2016-04-08 at 21 49 38

tbh I don't quite get what function we're after here. closing on the close button, for me on os x, never quits the app. it just closes the window. this is how it works in current master.

what are we trying to achieve? quitting on close?

boekkooi:
@algesten The reason for the PR is to allow any system to close to the tray not just OS X. 

I will take a look at the bug when I get my hands on a Mac but I think the problem is related to https://github.com/yakyak/yakyak/blob/master/src/main.coffee#L54

algesten:
i thought that the semantics on windows is that close button quits the app and minimise sometimes minimises to tray.

boekkooi:
@algesten Well I'm not sure about windows since I'm on gnome but what I remember is that more then one application has the option to minimize to tray on close (I think msn or live messenger had this option I'm sure ICQ did) and that a quit button on the tray icon or in the app would actually quit the application.

boekkooi:
@algesten I fixed the mac issue. The PR should be good now.

PR #156, opened 2016-03-24
Pull messages sent by self to the right

ppannuto:
This matches with the look that hangouts, iMessage, FB chat and
others use that I think most have become accostomed to. I find it
aids reading back through a chat.

Other style notes/changes:
  - Add a slight background color to messages sent by self
  - Round off grouped messages such that they appear as one block
  - Remove name for messages sent by self (redundant, looks funny IMO)
  - Chat heads stay in the left column for quick scanning

I've never used coffescript or less before so if I did anything foolish, feel
free to yell at me and I'll fix.

screen shot 2016-03-24 at 01 49 56

### DIFF
--- a/src/ui/views/messages.coffee
+++ b/src/ui/views/messages.coffee
@@ -53,6 +53,7 @@ groupEvents = (es, entity) ->
                 cid: cid
                 event: []
             }
+        e.group_index = user.event.length
         user.event.push e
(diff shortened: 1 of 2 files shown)


algesten:
I expect #160 to be rolled into master in some shape or form. Closing this for now.

PR #166, opened 2016-04-09
Only allow a single running instance

boekkooi:
Hi there,

When using YakYak in gnome I have had the problem a couple of times that I started yakyak multiple times because I forgot that it was in the tray.
This PR fixes that by adding [`makeSingleInstance`](https://github.com/electron/electron/blob/v0.37.4/docs/api/app.md#appmakesingleinstancecallback) in combination with [`mainWindow.show`](https://github.com/electron/electron/blob/v0.37.4/docs/api/browser-window.md#winshow).

Cheers,
Warnar

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -45,6 +45,15 @@ seqreq = require './seqreq'
 
 mainWindow = null
 
+# Only allow a single active instance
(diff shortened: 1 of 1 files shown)


algesten:
thanks!

PR #188, opened 2016-06-08
Update to electron 1.2.1

fbaew:
Fixed incorrect casing in login.coffee which caused very annoying behaviours

1. Couldn't log in
2. Every 3 seconds a new login window would open :)


davibe:
thank you :)

PR #193, opened 2016-06-09
fix an issue with jQuery global namespace registration in Login

Senney:
Some environments redirect to a third-party site when authenticating to Google (using SAML). In my case, the third-party site had several jQuery plugins implemented which required window.jQuery to be defined. 

See this issue for more information:
https://github.com/electron/electron/issues/254


Senney:
Accidentally committed with the wrong ID :^) I'll re-open another pull request with the correct git information.

Issue #192, opened 2016-06-09
Electron Process doesn't close on X in windows


AradonArmadon:
This was done on Windows 10 64-bit.

I downloaded the latest binary (1.2.3) and started up yakyak.  I then closed it via the 'X' at the top when the sign in with your google account screen came up.  After that, I tried to open it again and it complained that a lock file was currently in use by another electron process.

After some digging, I found that the Electron process was still running in the background.  After killing the processes, I was able to launch yakyak again.

It should be noted, that going to the menu and hitting quit does close the process correctly, it is only when the X is hit.

After logging in fully, it looks like the app correctly closes upon hitting X, so this may only be an issue when doing the oauth via google.

davibe:
I am not sure about your last paragraph because you say that x closes correctly + some oauth thing.

Anyway. We have 0 windows developers / contributors at the moment therefore it's difficult for us to fix window-specific bugs. I mark this as 'help wanted'. If i find the time to fire up a vm i will have a look but it's possible i won't do it :P

AradonArmadon:
The last paragraph oauth thing is in reference to the login.coffee script.  Specifically you can see the oauth reference here: https://github.com/yakyak/yakyak/blob/master/src/login.coffee#L18

When I say after the X works correctly, I mean it only works correctly after you've logged into google and are in the app proper.  While you are performing the oauth login to google, the electron process does not close if you hit X (and actually locks you out of the app until you are able to kill the process).

I understand you may not have current windows contributors, but it never helps to get these things documented as known issues :)

davibe:
I think i got it. While you log-in you use a separate window. The main window is "hidden". If you manage to close the login window the app sits in the background because there is still a hidden window that is never closed. I may be able to work with this even if i don't have windows.

Senney:
... Clearly you can see I was having a rough time with my repo. Pull request created to fix this issue.

PR #195, opened 2016-06-09
fix app to exit gracefully when login window closed (#192)

Senney:
Closing the login window before successfully authenticating
will now close the application as expected.

Fix issue #192


davibe:
Great, thank you sir! :)

PR #217, opened 2016-06-16
You gotta call things what they are. ;)

eliduke:
Thanks for making such an AWESOME bit of software.


davibe:
Thank you for your enthusiasm. I just don't want to loose modesty

PR #231, opened 2016-06-17
Issue 230

mattselph:
Issue 230; Now only prompts a user to upgrade when version is less than released version.

### DIFF
--- a/src/ui/version.coffee
+++ b/src/ui/version.coffee
@@ -14,9 +14,9 @@ check = ->
     releasedVersion = tag.substr(1) # remove first "v" char
     localVersion = require('electron').remote.require('electron').app.getVersion()
     versionAdvertised = window.localStorage.versionAdvertised or null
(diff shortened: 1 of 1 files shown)


davibe:
Since you're comparing strings such as `"1.2.1"`and  `"1.2.2"` I am not sure your thing works in all cases (it's one of those things i always get confused about)

That being said you also removed the check that checks if we have already notified the user for that specific new version. So you would end up notifying the user each time he restores the window.

arnriu:
Agree with @davibe , pre-release are pre-release... Maybe we could identify pre-release by appending a letter (1.3.1-a) and test upgrade only on major version without any letter ?

davibe:
@arnriu i don't know. Having a non-number would make it more difficult to check versions.

Anyway. Better way to check versions would be to convert it to int and see which one is actually higher, then only notifying the user once for every new version. At the moment this patch is wong.

I can do this in my next commit

Issue #229, opened 2016-06-17
Reflow text for narrower windows


wigster:
Currently (Yakyak 1.3.0, Windows 10), when the window is made narrower than ~1000 pixels the right edge of the conversation pane disappears. 

Would it be possible to squeeze the empty space between the right and left chat bubbles instead of scrolling it off? Or is that what Google sends from its servers and cannot be altered?

davibe:
can you provide screenshots ?

wigster:
[https://1drv.ms/i/s!Aq44GeJ_w3olht1TuGlJ-US3Z0X3Ew](Try this.)

wigster:
[Try this.](https://1drv.ms/i/s!Aq44GeJ_w3olht1TuGlJ-US3Z0X3Ew)

This is the full window, when I shrink it horizontally.

davibe:
oh my. That should not happen. It does not happen on other platforms

wigster:
In case it has some sort of impact, I am running on a 4k screen with 150% DPI scaling.

ZackBoe:
I'm also seeing this, but only occasionally, on Debian 8.

I've run into a couple somewhat related issues as well with the otherwise wonderful new design:
- Absurdly long group names (dont ask) should be truncated somehow
- Conversation list now shows less on screen, doesn't fully hide text when shrunk to its minimum

Lemme know if you want me to open separate issues for either of these.

All three issues are present here:
![0e1c1fc3](https://cloud.githubusercontent.com/assets/904055/16162456/a1fb7818-34a1-11e6-97ba-ebc84f8547bf.png)
For comparison:
![c6bdabf1](https://cloud.githubusercontent.com/assets/904055/16162474/b938aa00-34a1-11e6-9c45-43a08efacffa.png)

davibe:
for the moment you don't need to open separate issues. I just want to raise attention of @arnriu. Mabe he has some ideas.

arnriu:
@davibe Yes sir, I have ;p

davibe:
Why would those 2 lines fuck up things?

arnriu:
Those 2 lines tells that the view has been updated and, among other things (DOM manipulation, models update, ...), redraw the whole applayout. It's very heavy, and we do not need to do that on resize nor move.
Well, one can say, we do not need to do that, but what's the big deal? We call this often. When we change current conversations, add new one, etc... Even on app start. And it never hang the whole application.

[...]

Issue #216, opened 2016-06-16
'Conversations' stomped by 'Max' button


dliscomb:
YakYak 1.2.1 (1.3.1)
On OS X 10.11.15, the title bar text 'Conversations' gets stomped by the Close/Min/Max buttons:

![image](https://cloud.githubusercontent.com/assets/10132519/16117628/0b4b8bf2-33a1-11e6-8b6b-752da1e28f10.png)

arnriu:
Try ⌘+ or view -> Zoom in.

Will look into it later to make this part of window responsive to zoom.

randymorris:
I can confirm that zooming out causes this, but the default zoom level displays correctly.

Issue #211, opened 2016-06-16
Hide window borders on Windows


Naatan:
On windows it shows the native window borders (including titlebar).

davibe:
we need windows contributors for this. I am not a windows user.

tvthatsme:
Confirmed that windows shows the native borders which takes a lot away from the sleek look.

The route forward for this would be to add a frame property if the process.platform is 'win32' to the creation of the browser window (src/main.coffee):

```
    mainWindow = new BrowserWindow {
        width: 730
        height: 590
        "min-width": 620
        "min-height": 420
        icon: path.join __dirname, 'icons', 'icon.png'
        show: true
        titleBarStyle: 'hidden-inset' if process.platform is 'darwin'
        frame: false if process.platform is 'win32'
    }
```

However, this removes the toolbars as well (YakYak, View, Color Scheme). So this change is going to need to dig deeper into the UI of the app to be able to expose the toolbar menus without the frame support. I'd like to help but can't promise my availability for windows access.

Just wanted to help get that conversation started :smile:

Naatan:
I guess it needs a burger menu or something similar. Or you can just make it an ALT menu (hide it by default, show it when someone presses ALT). This would be an issue on linux as well.

davibe:
Maybe we could un-hide borders when mouse reaches the top yakyak titlebar ?

Naatan:
That sounds really awkward. Imo ALT hiding would be a better/cleaner solution.

See:

https://github.com/electron/electron/blob/master/docs/api/browser-window.md#new-browserwindowoptions

Ctrl+F for `autoHideMenuBar`

tvthatsme:
Going the ALT or mouse proximity methods hides information from the users which is normally bad from a usability point of view. You have to know that functionality is there before you can use it.

I would favor a full integration of the toolbar into the app (think <header>) across all platforms. It would be a lot more work but it could be beautiful and wouldn't put the app at Window's or Linux's mercies on how things get styled. It would affect a lot more people though so idk if its worth all the trouble.

tvthatsme:
Or... hamburger menu on the top right?

Naatan:
Couldn't it just be added here?

![image](https://cloud.githubusercontent.com/assets/278649/16161304/c49e0536-349a-11e6-9506-3f90ef4ab009.png)

tvthatsme:
Yeah, that's exactly what I was thinking!

davibe:
[...]

PR #236, opened 2016-06-17
On resize / move , no update viewstate

arnriu:
#229 Windows performance issue

It seems like we do not need to update viewstate on these events.
These events fire A LOT when you resize and can hang or crash the app if you do heavy stuff like updating viewstate.

Tested on windows 10, i7 6700k + 32gb ram. Before, app took ~3secs to update view when I tried to resize, and sometimes froze or crashed. Now, it's so smooth :)

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -109,6 +109,7 @@ app.on 'ready', ->
         icon: path.join __dirname, 'icons', 'icon.png'
         show: true
         titleBarStyle: 'hidden-inset' if process.platform is 'darwin'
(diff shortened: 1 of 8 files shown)


arnriu:
https://github.com/yakyak/yakyak/issues/211
Not perfect, but rewrite the whole menu would be very long. For now, I just added hidden menu, and a button to show it (pressing alt will show menu too).

https://github.com/yakyak/yakyak/issues/216
Issue addressed, need testing.

PR #239, opened 2016-06-18
Hide window on ESC key

oezdemir:
Alternative shortcut to hide the window. It is not practical every time to use the mouse to hide. Using the ESC key is a fast and intuitive way to hide the window quickly.

### DIFF
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -109,6 +109,10 @@ handle 'toggleshowtray', ->
 handle 'togglehidedockicon', ->
     viewstate.setHideDockIcon(not viewstate.hidedockicon)
 
+handle 'hideWindow', ->
+    mainWindow = remote.getCurrentWindow() # And we hope we don't get another ;)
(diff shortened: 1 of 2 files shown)


davibe:
- this only works with focus on the input area. while focus is there most of the time the shortcut shold be associated to the entire app
- on mac you can do cmd + h or cmd + w. Other OS may altread have their own shortcuts i think, isn't it ?

bryanzw15:
On Windows/Linux you can use ctrl+h to hide the window

ctcpip:
@bryanzw15  -  CTRL+H does nothing on Linux (LXDE at least)

bryanzw15:
@ctcpip I just tested on Windows 10 and Ubuntu 16.04 and it works, maybe you're not fully updated? Otherwise it's just a weird issue with LXDE

ctcpip:
More likely just the difference in desktop environment than anything else.

davibe:
I'm not sure if this ESC key is a good idea. I merge it for the moment

Issue #241, opened 2016-06-19
Hamburger menu breaks Edit menu functionality for Windows


bryanzw15:
Having added the Edit stuff the other day, the new addition of the hamburger menu actually breaks the cut, copy, paste, and select all options on there. This is because when the hamburger menu is clicked, the focus is taken off of the text area (the cursor leaves) and it doesn't allow it to work. If redo or undo is clicked and the focus is returned to the text area, they work properly again.

If there is some way to maintain focus on the text area when the hamburger menu is clicked, I think that would work, though I don't know how to do that. The only other option I can think of is just removing those menu items again, even though I just added them, because they are fairly common keyboard shortcuts that work without the menu items there, I just thought it would be nice to have the same options as the Mac version does.

I'll look into maintaining the focus on the text area but if anyone thinks it better to just remove them, or has a better idea, I'll do that instead.

arnriu:
You can still access menu with "alt" key and use copy / cut / paste.
But I agree with you, those menu items are not really essentials nowadays. If you look at windows apps, like the mail app, copy / cut / paste are not in menus.

bryanzw15:
I feel bad removing them because I just added them like a day ago but it seems like the best solution for now. I'll issue a pull request to remove the broken ones in a bit.

PR #243, opened 2016-06-20
Remove broken edit options

bryanzw15:
#241 Removing the edit options (cut, copy, paste, select all) that were broken with the addition of the hamburger menu.

Side-question: I've not used GitHub/Git all that much, how would I make it so all my other commits don't show up at the bottom? Just the relevant one (remove broken edit functionalities)


marcusmotill:
Using the github desktop client you can combine commits into one big one.  Thats not really good practice unless you have commits only to switch branches and look at those.  So keep all commits unless they are half done really

PR #248, opened 2016-06-20
shortcode emoji support

acordar:
Added the ability to use emoji shortcode ( :smile: ) as a shortcut to
all of the emojis. Right now this only supports the emojis from the
People emoji category but can be expanded. Based on shortcode from
http://www.emoji-cheat-sheet.com/


jadengore:
+1 to this, thanks @acordar for the work! A feature I've been missing!

TadeasKriz:
Would it be possible to allow more aliases? For example the `:smile:` could be represented by `:)` or `:-)` depending on how people write smileys. I personally like the way hangouts smileys can be typed https://aprescott.com/posts/hangouts-emoji so alternative aliases would be awesome! :)

acordar:
@TadeasKriz, my latest change as added what you requsted

TadeasKriz:
Wow! Thank you, that's stellar!

acordar:
I think I have added an emoji shortcode for all of the emojis now. A neat feature to consider also would be having an autocomplete feature with previews of the emojis as it's pretty much impossible to remember all of them. If you've used slack, they have an autocomplete when using emoji shortcodes.

jadengore:
@acordar This might be a good v1, and we can open an issue that adds an autocomplete. This is definitely a good MVP for a better emoji experience in YakYak.

@davibe What do you think of this?

davibe:
allright

jadengore:
@acordar I've noticed a couple emoji still dont line up to the shortcodes, but maybe we can add some more as time goes on. Good work.

naydichev:
I think this is causing some URLs to have emojis... I'm seeing the ```https://``` be converted to ```http😕/```.

PR #262, opened 2016-06-23
Added ability to change app font size

marcusmotill:


arnriu:
Why would you need this over the zoom functionality ? (no hard feelings, I'm just trying to understand what it is useful for :) ).

marcusmotill:
Sorry not familiar with the zoom functionality!  Can you elaborate?

marcusmotill:
Just looked at it now.. I see how they overlap.  Seems like the zoom zooms everything, this just does font size.  idk a bit redundant yes.  Adds some more functionality 👍

marcusmotill:
Either way I need to make it so the font size isn't getting reset on zoon

arnriu:
Yes, zoom scale the whole ui by a factor (if you zoom in one time, the whole ui will be 125% bigger, including font).

marcusmotill:
Zoom feature is kind of annoying in my opinion.  To get the font to a
readable size for me the UI gets jammed with oversize icons and margins.

monchote:
I like this feature as the current font sizes are not very well balanced in my opinion and zooming the whole UI is not the right solution.

PR #269, opened 2016-06-25
twitter preview

bseber:
I thought it would be nice to have a twitter preview :-)

I've refactored the format section a little bit. I think this makes it a lot easier to add more renderers like twitter and youtube preview for instance.

Please note that the format refactoring results in rendering image links as plain image now. The link is rendered above as anchor tag. Personally I prefer seeing urls, so... But we can merge the image and text rendering again, of course.

Feedback welcome!

### DIFF
--- a/src/ui/css/yakyak/messages.less
+++ b/src/ui/css/yakyak/messages.less
@@ -92,6 +92,12 @@
                         color: var(--darkgrey);
                         font-size: 0.9em;
                     }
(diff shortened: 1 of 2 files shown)


davibe:
After accepting other pull requests there are some conflicts. Can you please rebase your work ?

bseber:
@davibe done

PR #271, opened 2016-06-25
Enable spellchecker

monchote:
Only works for the system language and underlines misspelled words. It doesn't offer suggestions at the moment.

Fixes #61

PR #276, opened 2016-06-26
Enable spellchecker

### LINKED ISSUE #61: Spell check as you type

monchote:
Highlights misspelled words but doesn't offer suggestions at the moment.

Fixes #61
Relates to #271

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "hangupsjs": "^1.3.1",
     "moment": "^2.10.3",
     "node-notifier": "^4.2.3",
+    "spellchecker": "^3.3.1",
     "notr": "^1.1.2",
     "q": "^1.4.0",
     "request": "^2.67.0",
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -40,11 +40,15 @@ handle 'init', (init) ->
     require('./version').check()
 
     require('electron').webFrame.setSpellCheckProvider navigator.language, true, spellCheck: (text) ->
+        isMisspelled = false
         for dictionary in spellchecker.getAvailableDictionaries()
             spellchecker.setDictionary(dictionary)
             if spellchecker.isMisspelled(text) != false
-                return true
-        return false
+                isMisspelled = false
+                break
+            else
+                isMisspelled = true
+        return !isMisspelled
 
 handle 'chat_message', (ev) ->
     conv.addChatMessage ev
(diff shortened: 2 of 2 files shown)


monchote:
Make sure to clean the node modules before compiling and running:
```
rm -fr node_modules/
npm install
```

davibe:
ok i try again

davibe:
I am still getting the same error

monchote:
@davibe What version of node do you have?

davibe:
4.4 but i think node has little to do with this

davibe:
dade@choo:~/work/yakyak(master)$ node -v
v4.4.4

monchote:
I'm on v6.2.2 and not getting that error.

What do you suspect it can be?

davibe:
I have no idea. Maybe the module is built with a node-gyp version that is different from the electron one

https://github.com/electron/electron/blob/master/docs/tutorial/using-native-node-modules.md

davibe:
I have reverted the merge using the --hard way. I hope nobody pulled this in the mean time

monchote:
@davibe I've just tested it on 2 more dev environments and works just fine so I suspect the problem is there's something different in our node/npm setups.

More than happy to help you figure out what it is as I think this feature is incredibly important.

See screenshot attached of it working.
screen shot 2016-06-27 at 10 51 37

davibe:
I understand. I don't have time at the moment. If it was something i could just merge and go then ok. Otherwise i will have to check with more calm. 

I am also thinking that since i build the windows/linux binaries on osx a native module will have similar problems on other platforms ...

monchote:
@davibe [node-spellchecker](https://github.com/atom/node-spellchecker) works with Windows, macOS and Linux (via Hunspell). I'm installing VMs to test it thoroughly in Linux and Windows now.

davibe:
I know it works as a npm package. But if it's BUILT on mac it may not work on other os

monchote:
Oh, I see what you mean. I'll test it and report back.

monchote:
@davibe So the binaries built from mac don't work on Windows or Linux. Managed to build from Linux and the spellchecker works. Building from Windows is a total nightmare.

davibe:
yesterday i tried looking for `npm search spellchecker` and I think there may be some other alternative that does not require native building. Some of them are specifically targeting electron. No idea which one actually works though.

monchote:
@davibe Thanks, I'll keep looking into it when I have more time. Shall we reopen #61 in the meantime?

Issue #297, opened 2016-06-30
Multi user chat name labels


a1dutch:
When having multi user chats, its hard to tell who has typed what without hovering over the persons icon. its alot easier if the user has set an picture, but when there is no picture there needs to be some way of displaying the name of the person next to what they have typed.

davibe:
@arnriu

randymorris:
If you can't find a nice way of displaying the whole name maybe you could do what Mail.app does on OSX if a contact doesn't have an image.  Rather than a placeholder image it displays their initials:

![screen shot 2016-06-30 at 8 38 28 am](https://cloud.githubusercontent.com/assets/17761/16488340/0c1b2f4e-3e9e-11e6-881f-bc6023861b09.png)

a1dutch:
any idea when this will be available @arnriu ?

a1dutch:
just seen your pull request, thanks

arnriu:
@a1dutch You're welcome :smile:

PR #299, opened 2016-06-30
Multi user chat name labels #297

arnriu:
User initials when no profile picture available.

initials


arnriu:
capture d ecran 2016-06-30 a 17 41 18

arnriu:
emoji

ADetectionAlgorithm:
I would recommend that these changes go through.  They are nice and necessary (especially emoji on windows)

PR #314, opened 2016-07-09
Select Conversation <Number> Menu Items and Shortcut Keys

rajington:
`Select Conversation #` shortcut keys match https://github.com/atom/atom/blob/master/keymaps/win32.cson#L85 and popular browsers.

macOS `Previous Conversation` and `Next Conversation` shortcuts keys match `Message.app`

### DIFF
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -99,6 +99,11 @@ handle 'selectNextConv', (offset = 1) ->
     viewstate.selectNextConv offset
     ipc.send 'setfocus', viewstate.selectedConv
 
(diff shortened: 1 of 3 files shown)


davibe:
I don't think Messages does alt+number to select convs. Do we want that ? @arnriu @maxkueng @anyone ?

rajington:
It's useful for me, but I don't take it personally if it's not in master. Messages doesn't have that one, but it doesn't really have favorites. I pin tabs in Chrome, and use those keys to get to the pinned ones.

davibe:
ok, the favourites are a good point for this. Other opinions ?

PR #319, opened 2016-07-15
Ctrl and enter to send

easauceda:


PMiner:
Yay! Thanks!

easauceda:
No problem!

I noticed that just enabling ctrl + enter would disallow from using it to add line breaks to your message, so I think this strikes a good balance for people who prefer one behavior over another.

PMiner:
Perhaps shift+enter for break?

easauceda:
can't believe I missed that...good catch! really makes it a lot less complicated

davibe:
So basically you want to be able to do ctrl + enter because when you copy-paste and send stuff you forget to release ctrl ? Am i getting this right ?

davibe:
So basically you are adding a toggle in the menuitem + adding a new shortcut (shift+enter) just because you forget to release ctrl before hitting enter. I am not sure about this, hmm..

conceptualspace:
why not just be consistent with hangouts? thus shift+enter = break, and enter/ctrl+enter = send

PR #320, opened 2016-07-20
Added monochrome tray icon (and high dpi) for the OSX

Stanko:
Hi guys, thanks for the hard work!

I added separate tray icon to match overall OSX styling (monochromatic). As I'm not that familiar with coffeescript, please check it before merging in. There are also "original" svgs for the mono icons.


davibe:
i am thinking what happens on windows / linux ?

Stanko:
@davibe I've added a check for OS, so Windows and Linux are still using the current icons for the tray.

davibe:
right, ty

PR #322, opened 2016-07-21
Re-add retina tray icon

monchote:
I added this for the yak icon and it got lost when the new icon got reintroduced. 

This is different from #320


monchote:
@davibe Merge conflict resolved.

davibe:
@monchote is this still relevant after https://github.com/yakyak/yakyak/pull/362 ?

monchote:
@davibe Yes, we still need this as the other menu icon with retina versions is only for macOS. This is the one shown in Linux and Windows(?) which also support hi-dpi screen these days.

davibe:
thank you

PR #329, opened 2016-07-25
only renders emoji if followed by space or enter

averissimo:
Better replicates Google Hangouts emoji conversion by:
- Only converting emojis that a whole word
- Converts last written word to emoji if press enter

Test cases (spaces or enter are denoted as ``«space|enter»``)
 "I«space»like«space»it!«space»:)«enter»"  -> correct emoji
":D«enter»"  -> correct emoji
"-_-«space»«enter»" -> correct emoji
"The«space»url«space»is«space»http://github.com«space»«enter»" -> "The url is http://github.com"
"The«space»url«space»is«space»http://github.com«enter»" -> "The url is http://github.com"

selection_022

### DIFF
--- a/src/ui/util.coffee
+++ b/src/ui/util.coffee
@@ -82,16 +82,21 @@ convertEmoji = (text) ->
     unicodeMap = require './emojishortcode'
 
     patterns = [
-        ":[a-zA-Z0-9_\+-]+:",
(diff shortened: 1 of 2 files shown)


kolorafa:
Can you disable converting to emoji? (and personally i think emoji should be changed as you type, so you know what you send, but i don't like emoji because most of them are total crap... or even fail to render on a lot of clients)

averissimo:
Super valid point, if they didn't render on purpose (when pressing enter) then I guess changes to  ``src/ui/views/input.coffee`` should not be included.

I included it because that is the behavior with hangouts.

averissimo:
the option would be a great addition, but I think that should be implemented in a different pull request as new feature/enhancement

averissimo:
it was actually pretty easy to implement, here is a branch synced with yakyak/yakyak

https://github.com/averissimo/yakyak/tree/emoji_option

I guess there should be an issue to discuss this

dhensche:
I pulled down the branch listed above (emoji_option) and when I paste the following URL into the message box and hit enter it is turning the ":/" into the corresponding emoji -- http://arstechnica.com/science/2016/08/dentists-forgot-to-study-flossing-for-a-century-recommended-it-anyway/

PR #341, opened 2016-08-17
corrects cursor position when hitting space (if it has emoji short code in the msg)

averissimo:
This fixes the cursor position to stay as it was before hitting space.

This is relevant when there is an emoji on the message (eg. ``:)``):

- editing something in the middle of the message and pressing space would move the cursor to the end of the message

### DIFF
--- a/src/ui/views/input.coffee
+++ b/src/ui/views/input.coffee
@@ -110,8 +110,14 @@ module.exports = view (models) ->
                 #check for emojis after pressing space
                 if e.keyCode == 32
                     element = document.getElementById "message-input"
+                    # get cursor position
+                    startSel = element.selectionStart
+                    endSel  = element.selectionEnd
                     # Converts emojicodes (e.g. :smile:, :-) ) to unicode
                     element.value = convertEmoji(element.value)
+                    # Set cursor position (otherwise it would go to end of inpu)
+                    element.selectionStart = startSel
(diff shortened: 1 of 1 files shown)


davibe:
Are you using osx ? because i can't reproduce. I can edit message in the middle, hit space, and the cursor does not jump to the end of the message.

randymorris:
I can reproduce.  This only seems to come into play when there is an emoji short code at the end of the line that hasn't been replaced yet, e.g. (with `|` as the cursor):

    The quick brownfox jumps over the lazy dog :)|

Move the cursor to the middle of the line:

    The quick brown|fox jumps over the lazy dog :)

Press space to fix the typo and the cursor jumps after the emoji:

    The quick brown fox jumps over the lazy dog 😄 |

averissimo:
In my system (linux with yakyak/yakyak/master checked out) it always moves to the end (with ``|`` as cursor)

    The quick brownfox :)|jumps over the lazy dog

Press space

    The quick brownfox 😃 jumps over the lazy dog|

randymorris:
Indeed that also triggers it.

averissimo:
ohhhh, my bad!! this is specific to emojis, it was so obvious when I was writing the pull request that I forgot to write it in the title/body

davibe:
I trust you, i merge. I am too lazy to try :)

Issue #349, opened 2016-08-25
Ability to only show user picture


zachthezman:
First off, top-notch job here. Love this app. I use it all the time.

I would love to see an option to only show the user's picture in the left pane. You've already added the ability to hide the timestamp and last message. My goal is to slim up the left pane to be as minimalistic as possible. I split the app with Slack on my screen so the more real estate I can free up in the actual conversation pane on the right would be great.

randymorris:
This used to be possible just by resizing the pane.  I'm guessing this changed when the overall theme changed a while back.  Here's an old screenshot from the old v1.0.0 README that showed this:

![Screenshot](https://cloud.githubusercontent.com/assets/227204/8255540/d922d252-16a0-11e5-86b2-bfec901bbdbc.png)

arnriu:
I don't know if I can help with this... Maybe if you could grab me a beer...

Ok, ok, fine. Here you are: #350

zachthezman:
![Mother of God](http://i.giphy.com/kCo3euE2yScYo.gif)

@arnriu you are a rock star! Thank you.

averissimo:
already implemented and merged

PR #350, opened 2016-08-25
Minimal convlist option, unread counter style.

arnriu:
#349

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -27,7 +27,8 @@ paths =
     lessd:   './src/ui/css/**/*.less'
     css:     './src/**/*.css'
     fonts:   ['./src/**/*.eot', './src/**/*.svg',
-              './src/**/*.ttf', './src/**/*.woff']
+              './src/**/*.ttf', './src/**/*.woff',
+              './src/**/*.woff2']
 
 # setup package stuff (README, package.json)
 gulp.task 'package', ->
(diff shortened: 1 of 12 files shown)


arnriu:
capture d ecran 2016-08-25 a 16 09 21

davibe:
@arnriu i believe for some user like me "minimal" would be just the text (contact name). Can you think of a better menu arrangement/name ?

Also. Is "animated thumbnails" compatible with "minimal" mode ?

arnriu:
@davibe I know, the text isn't perfect, I was hoping for some help, since english is not my native language... ;)
I tried a new menu arrangement, let me know what you think about it.

And yes, animated thumbnails are compatible.

arnriu:
capture d ecran 2016-08-29 a 17 08 20

davibe:
Since it's all "show" we could put "show animated thumbnails". I like this version better. I am not native english either, but who is :)
I will test it soon, i hope

randymorris:
I haven't had a chance to test this, but does unchecking the first option disable the other two?  I would think that it should since only thumbnails and animated thumbnails sort of contradict not showing any thumbnails.

arnriu:
@randymorris no, if you uncheck first option, second and third options are still available:
- if you check second option while first is unchecked, it will check it. If you uncheck first option while second is checked, it will uncheck it.
- animated thumbs are also used in messages, so it is not dependent on conversation list thumbs.

Maybe the text are wrong? I'm open to any suggestions :)

randymorris:
Ah, I forgot about the message pane.  Forgive me.

arnriu:
capture d ecran 2016-08-30 a 15 57 15

Issue #331, opened 2016-07-27
Should there be an option to convert to emoji as you type?


averissimo:
This is a question that came from pull request https://github.com/yakyak/yakyak/pull/329

I have implemented this option in a0a0f2507d02a76692a42af1db9c0dc4da4446e4

It works with minimal (manual) tests

Feedback?

notdevinclark:
Thank you for implementing this option. Was looking for it all over the place in the stable release.

davibe:
i guess we can close this

PR #360, opened 2016-08-30
#331 Option in me enabling/disabling converting to emojis in messages

averissimo:
#331 Option to enable/disable convert to text to emoji

image

*Possible conflict* with #319 which is a much cleaner implementation on sending messsages.

### What does it do:

When disabled ':)' and other emojis are not converted to unicode and are just sent as plain text.

### What it does not do:

Convert incoming or past messages from emojis to text.

### Example:

- Message one - enabled
- Message two - disabled

image

Is there a convention on option titles? (almost all are all capitalized, except "show system tray"

### DIFF
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -359,6 +359,9 @@ handle 'showconvtime', (doshow) ->
 handle 'showconvlast', (doshow) ->
     viewstate.setShowConvLast doshow
 
+handle 'convertemoji', (doshow) ->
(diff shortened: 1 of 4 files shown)


averissimo:
@yakyak @davibe if i'm jumping ships on creating pull requests and not discussing this on issues let me know :)

I don't want to waste people time with unnecessary features/fixes just because I can code :)

davibe:
I prefer PRs. There is plenty of room to discuss here. 

No conventions on menu items we just try to make all of them similar.

Issue #302, opened 2016-07-02
Support Multiple google accounts


eyadsibai:
I have two different gmail accounts that I would like to login with.

invisiblegarage:
Agreed. Great app. Would love support of mulitple google logins.

davibe:
we already talked about this. It require quite some effort but it's interesting.

jlp78:
This is a duplicate of https://github.com/yakyak/yakyak/issues/13

davibe:
closing as duplicate

Issue #13, opened 2015-06-22
Multiple Accounts


piotrromanowski:
Implement the ability to log into multiple accounts

Note: This is a game changer

algesten:
This requires someone to investigate how a "standard" Google client logs into multiple accounts and keep the events and requests separate. It's on the hangupsjs level.

davidbanham:
The other approach would be to run a separate instance of hangupsjs for each account.

algesten:
and juggle a bunch of cookie stores. hm. could be done. there probably is a better way though...

the actual "logging in" to multiple accounts is likely to be straightforward. hangupsjs already contains code that does the initial oauth, followed by something called an "uberauth" (love the name), followed by a "merge session".

* oauth: https://github.com/algesten/hangupsjs/blob/v0.9.1/src/auth.coffee#L85
* uberauth: https://github.com/algesten/hangupsjs/blob/v0.9.1/src/auth.coffee#L173
* merge session: https://github.com/algesten/hangupsjs/blob/v0.9.1/src/auth.coffee#L185

as I understand it oauth establishes that you are you. uberauth is something that gives the established identity access to *everything google* https://github.com/tdryer/hangups/issues/123 which isn't really desirable but the only way we know about for now.

it wouldn't surprise me if you can uberauth into multiple accounts and then merge session them all to one single session.

the responses from google always contain "sender" information, so knowing which of multiple logins is receiving is not a problem.

each request is passing along the `client.id` as part of the payload, so saying which we're sending as is probably also straight forward. https://github.com/algesten/hangupsjs/blob/v0.9.1/src/client.coffee#L185

LukeAskew:
+1 Please do this.

pengi:
+1 for this too

pengi:
Is there any way of working around this issue for now? Like having two copies of yakyak, and just change a small parameter?

umidbekkarimov:
+1

danpritts:
+1

niftylettuce:
+1

freekrai:
+1

diego-pacheco:
+1

HarlemCake:
+1

brenna:
+1 i'd also be happy for now with some type of workaround involving two instances of yakyak

codeandrop:
+1

AdrianoCahete:
+1

jaygoldman:
+1

[...]

PR #380, opened 2016-09-18
Create command line flag to enable simultaneous multiple logins.

ctag:
This is a response for https://github.com/yakyak/yakyak/issues/13 and https://github.com/yakyak/yakyak/issues/302.

It's a hack which will modify the path of userData depending on a command line argument `user=<dir>`. So now several instances of yakyak can run from different directories, with different accounts logged in. 

Multiple windows must be used. I understand this isn't the clean solution (multiple accounts, single yakyak instance) wanted by users, but it works.. for me at least.

To streamline usage, the yakyak.desktop file can be copied, and each can then be modified to include this argument.

```
#!/usr/bin/env xdg-open

[Desktop Entry]
Name=YakYak

Add
GenericName=YakYak
Comment=Electron App from git
Path=/home/my-username/git/third-party/yakyak/
Exec=electron app user=my_first_account
Terminal=false
Type=Application
X-GNOME-Autostart-enabled=true
Icon=/home/my-username/git/third-party/yakyak/src/icons/atom.icns
Name[en_US]=YakYak [my_first_account]
```


averissimo:
cool initiative ^^

this seems to be javascript instead of coffeescript

see http://coffeescript.org/ for more information

it really really helps clean up the code, for instance the ``for`` cycle turns to: 

    regex = /^user=(.*)$/i;
    for lineArg in cmdLineArgs
        match = regex.exec(lineArg)
        console.log("match: " + match)
        if match
            ...

**Additionally**, when creating the structure for "command line arguments", one should really include the help argument and the structure to easily add new arguments.

ps. I am just a fan of the project and a coder so take my comments lightly :)

ctag:
Thanks for the tips, averissimo. I've never worked with coffeescript before, so they're appreciated.

If there's support/acknowledgement from the developers for this being the solution to the "multiple accounts logged in" request, I would be happy to learn some proper coffeescript and make a simple CLI argument handling system.

Right now the hack is certainly just a proof of concept, but I don't know how else to stage this conversation than to offer a pull request.

davibe:
If you write this in a good way i will merge it. However, it's just a thing for developers who want to have fun with it. It does not solve any of the problems we need to solve in order to make this a user feature.

davibe:
I close this to clean up things. If you want to just reopen it

PR #391, opened 2016-09-23
Simplify menu template by removing duplicate entries

averissimo:
I think a change in how the menus are built is necessary, as it requires two copy of each new menu item, **if not with this approach please give feedback and I'll glady implement it** :)

Maintaining duplicate menu entries for OSX and Others could lead to (temporary) missing entries in menu -- when devs can only test in one platform -- and unnecessary duplicate work

### Tests

**I have (manually) tested in Ubuntu:**
- menus are showing
- menus functionality works
- shortcuts work (except for zoom in, all seem to work)

**I haven't tested the OSX and Windows directly**, the menus seem to render as they should in Ubuntu (with platform tricked to be OSX), but that's how far I've gone.

I can test the Windows environment on Monday, but the OSX will have to be someone else :)

### DIFF
--- a/src/ui/views/menu.coffee
+++ b/src/ui/views/menu.coffee
@@ -34,16 +34,16 @@ acceleratorMap = {
 }
 
 templateYakYak = (viewstate) ->
-    tmp = []
+    tmpl = []
     if os.platform() == 'darwin'
-        tmp = tmp.concat [
+        tmpl.push [
             { label: 'About YakYak', selector: 'orderFrontStandardAboutPanel:' }
             #{ type: 'separator' }
             # { label: 'Preferences...', accelerator: 'Command+,',
             # click: => delegate.openConfig() }
             { type: 'separator' }
         ]
-    tmp = tmp.concat [
+    tmpl.push [
         {
             label: 'Hide YakYak'
             accelerator: acceleratorMap['hideyakyak'][os]
@@ -52,14 +52,14 @@ templateYakYak = (viewstate) ->
     ]
 
     if os.platform() == 'darwin'
-        tmp = tmp.concat [{
+        tmpl.push [{
               label: 'Hide Others'
               accelerator: acceleratorMap['hideother'][os]
               selector: 'hideOtherApplications:'
         }
         { label: 'Show All', selector: 'unhideAllApplications:' }
         ]
-    tmp.concat [
+    tmpl.push [
       { type: 'separator' }
       {
           label: 'Open Inspector'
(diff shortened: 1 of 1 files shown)


averissimo:
Had a stupid mistake, correct YakYak menu now :)

Tested in Windows 7 environment

image

image

image

averissimo:
any preference between?

```
a.push.apply a, [...] # currently used to concat arrays in my implementation
a = a.concat [...]
# or other
```

src/ui/views/menu.coffee
```diff
@@ -1,74 +1,168 @@
...
-        { type: 'separator' }
-        # { label: 'Preferences...', accelerator: 'Command+,', click: => delegate.openConfig() }
-        { type: 'separator' }
-        { label: 'Hide YakYak', accelerator: 'Command+H', selector: 'hide:' }
-        { label: 'Hide Others', accelerator: 'Command+Shift+H', selector: 'hideOtherApplications:' }
-        { label: 'Show All', selector: 'unhideAllApplications:' }
-        { type: 'separator' }
-        { label: 'Open Inspector', accelerator: 'Command+Alt+I', click: -> action 'devtools' }
+acceleratorMap = {
+    hideyakyak:           {others: 'Control+H', darwin:'Command+H'}
```
davibe: I don't like putting things in columns like that. I know it kind of makes things more readable but 
- it makes it more difficult to edit things if you use more basic editors (i know there are ways..)
- it tends to make all your code wider which leads to trying to do strange things to avoid too long lines
- when you change something optimal column width may change and you have to tidy it up

Anyway.. it's no a problem. I can live with it, this is just a comment

src/ui/views/menu.coffee
```diff
@@ -1,74 +1,168 @@
...
+    conversation7:        {others: 'Alt+7', darwin:'Command+7'}
+    conversation8:        {others: 'Alt+8', darwin:'Command+8'}
+    conversation9:        {others: 'Alt+9', darwin:'Command+9'}
+    minimize:             {others: '', darwin:'Command+M'}
+    close:                {others: '', darwin:'Command+W'}
+}
+
+templateYakYak = (viewstate, platform) ->
+    tmpl = []
+    if platform == 'darwin'
```
davibe: Even if you created `acceleratorMap` you still end up doing `if platform ==` so now you have 2 ways to
 differentiate the platform.

src/ui/views/menu.coffee
```diff
@@ -1,74 +1,168 @@
...
+        ]
+    tmpl.push {
+        label: 'Hide YakYak'
+        accelerator: acceleratorMap['hideyakyak'][platform]
+        selector: 'hide:' if platform == 'darwin'
+        role: 'minimize' if platform == 'others'
+    }
+
+    if platform == 'darwin'
+        tmpl.push.apply tmpl, [{
```
davibe: Anyway using array.concat() is more `functional` style as it avoids mutations: `tmpl = tmpl.concat([..])`
Also know that you can do `array.push(array...)` in coffeescript.
Since I don't think efficiency is relevant here maybe you could also push each item one by one.
I would avoid using `apply`.

averissimo: lol cool feedback I'll make the changes :)

src/ui/views/menu.coffee
averissimo: yeah it kinda sucks, especially because there are shortcuts only for OSX mixed with "others"

I wanted to avoid having a lot  of if platform == 'darwin' spread throughout the menu generation (at least more than absolutely necessary)

I think I'll change the accelaratorMap to a function that has platform specific ('darwin') and default (now the 'others'... just realized the Lost reference :) ). This might clean up a little.

src/ui/views/menu.coffee
averissimo: I did it to make sure the mapping was correct and I didn't do something stupid copying and pasting.

I believe in conventions, so for me the argument that throughout the code it won't be like this wins me over :)

averissimo:
moved platform variable as global variable, allowing for specific shortcuts for all 3 platforms in the future

all the == 'darwin' was replaced by a variable called isDarwin, seems more sane (same with 'others' that no longer is used and replaced by isNotDarwin) -- reducing the number of '==' comparisons to just a boolean variable

can use if inside array, great tip @davibe :) although electron needs to filter out undefined entries to render menus

```
[
  {...}
  ...
  if ...
    {
        ...
    }
...].filter (n) -> n != undefined
```

PR #393, opened 2016-09-26
Add "show popup notifications" option to tray

averissimo:
I think this option should also be available in the tray icon.

If someone is concerned about the visibility of the messages, having to open the main window could be a problem.


davibe:
I believe this was already done in #452 right ? but it was done in a different way..

PR #395, opened 2016-09-26
Hide the window (close to tray) on all platforms

nerijus:
IMHO it should close to tray only if "Close to tray" is checked (smth like if global.closetotray).

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -371,15 +371,14 @@ app.on 'ready', ->
             ipcsend n, e
 
     # Emitted when the window is about to close.
-    # For OSX only hides the window if we're not force closing.
+    # Hides the window if we're not force closing.
     mainWindow.on 'close', (ev) ->
-        if process.platform == 'darwin'
-            if mainWindow.isFullScreen()
-                mainWindow.setFullScreen false
-            if not global.forceClose
-                ev.preventDefault()
-                mainWindow.hide()
-                return
+        if mainWindow.isFullScreen()
+            mainWindow.setFullScreen false
+        if not global.forceClose
+            ev.preventDefault()
+            mainWindow.hide()
+            return
 
         mainWindow = null
         quit()


averissimo:
Fix does not work if windows is closed with option disabled (in other words, if it is supposed to quit).

The interface option that controls whether it closes or not is: `viewstate.closetotray`

the problem seems to be that the event `beforeunload` (app.coffee) doesn't stop the `quit` event (main.coffee) when option `viewstate.closetotray` is true

I never quite understood what `global.closetotray` did

nerijus:
If I use if viewstate.closetotray it complains about viewstate not defined. If I add viewstate = require './ui/models', it complains about localStorage undefined. Do you know how to fix it?

averissimo:
Yup!! I kinda thought I had a fix in my keep_in_tray branch, but that was all crap :)

The problem is that the `close` event is called before `beforeunload` when closing the window (link: http://electron.atom.io/docs/api/browser-window/ )

**The only thing necessary for this feature to work is change the name of the event from 'close' to 'unload'**

This need to be checked by someone with a OSX, as the forcequit 'feature' needs to be tested :)

src/main.coffee
```diff
@@ -371,15 +371,14 @@ app.on 'ready', ->
             ipcsend n, e
 
     # Emitted when the window is about to close.
-    # For OSX only hides the window if we're not force closing.
+    # Hides the window if we're not force closing.
     mainWindow.on 'close', (ev) ->
-        if process.platform == 'darwin'
```
averissimo: i guess this needs to stay for macOSX users

src/main.coffee
```diff
@@ -371,15 +371,14 @@ app.on 'ready', ->
             ipcsend n, e
 
     # Emitted when the window is about to close.
-    # For OSX only hides the window if we're not force closing.
+    # Hides the window if we're not force closing.
     mainWindow.on 'close', (ev) ->
```
averissimo: close -> unload
http://electron.atom.io/docs/api/browser-window/#event-close

averissimo: line 375

nerijus:
Well, I specifically removed "if process.platform == 'darwin'" as I want to be able to close to tray on Linux. If this check is removed, closing to tray is possible.

averissimo:
Yeah, but it becomes impossible to "not close to tray"

**When you rename the event from 'close' to 'unload' it will close to tray** as your implementation, but will also quit if the option if off

The code that really handles that is in `app.coffee` in the 'beforeunload' event handler, but that javascript event is never reached because 'close' is called before and it quits.

nerijus:
Aha, I see it. I will close it and create a new pull request.

PR #397, opened 2016-09-26
Make hide the window (close to tray) work on all platforms if "Close …

nerijus:


averissimo:
It works when building from source, but not when I'm running a binary version of the same code

Can you confirm this?

nerijus:
It works, it closes to tray if Close to tray" is checked. If it is unchecked, closing yields error:
```
Uncaught Exception:
Error: Attempting to call a function in a renderer window that has been closed or released. Function provided here: trayicon.js:47:17.
    at Tray.callIntoRenderer (/opt/yakyak/dist/linux-x64/resources/electron.asar/browser/rpc-server.js:190:19)
    at emitTwo (events.js:106:13)
    at Tray.emit (events.js:191:7)
```

averissimo:
arghhhh you are experiencing the same  thing :(

I think this will correct it: https://github.com/averissimo/yakyak/commit/ef9597da0d72c7847dba99f1e8898b1b2e603ed5

Nevertheless, I think it makes sense to handle all events in app.coffee, maybe @davibe has some insight why it was in main.coffee

@nerijus can you test this? and maybe add to your PR if it works

nerijus:
I get blank screen now when starting app (even without your patch). I suspect because of too new Electron version. How do I use an older version?

averissimo:
you have to merge #403 (or my branch) if you are using yakyak/yakyak master. I screwed up on a PR with a typo on the menu.coffee that might be causing that

nerijus:
With #403 blank screen is fixed, and closing works as expected. I added your commit to the PR.

PR #398, opened 2016-09-27
Image preview when pasting or adding images via selector

averissimo:
This improves current functionality and is only implemented when pasting images.

It allows to paste image and send both the image and current text with "enter" or clicking the image.

output
### Open questions / TODO
- **Done** _see comment below_: How to do this with multiple images? Once we figure this one out, I guess it becomes mergeable
  - The image selector allows to upload multiple images and that is why It is not implemented yet
- How to allow pasting even when input is not selected
  - Should we have a system wide event handler for pasting that always  paste to conversation (if conversation is the active interface)
- I don't like the fact that this implementation requires a div over 'emoji container', but otherwise it would have a conflict with it
### Testing

I've done some testing, but need further :)

edit1: adds figure
edit2: mark first point of TODO as done and changed title

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
   "homepage": "https://github.com/yakyak/yakyak",
   "dependencies": {
     "autosize": "^3.0.8",
-    "electron-debug": "^1.0.1",
     "hangupsjs": "^1.3.2",
     "mime-types": "^2.1.12",
     "moment": "^2.10.3",
--- a/package.json
+++ b/package.json
@@ -31,8 +31,9 @@
   "homepage": "https://github.com/yakyak/yakyak",
   "dependencies": {
     "autosize": "^3.0.8",
+    "electron-debug": "^1.0.1",
     "hangupsjs": "^1.3.2",
-    "mmmagic": "^0.4.5",
+    "mime-types": "^2.1.12",
     "moment": "^2.10.3",
     "node-notifier": "^4.2.3",
     "notr": "^1.1.2",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   "dependencies": {
     "autosize": "^3.0.8",
     "hangupsjs": "^1.3.2",
+    "mmmagic": "^0.4.5",
     "moment": "^2.10.3",
     "node-notifier": "^4.2.3",
     "notr": "^1.1.2",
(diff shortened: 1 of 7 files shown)


averissimo:
While using YakYak I was sending a friend an image from disk and realise it does not allow to select more than a single file :)

A fast coding sprint later it is implemented and with a fallback if by any change multiple files are selected (default to previous behavior that would send them all)

averissimo:
Also added drag and drop visual support (actual implementation was already done)

outumy0pxgsvg

averissimo:
noticed it was lacking a pointer when hovering the send image button.

- adds pointer
- adds animation with color change of button (and back to normal)
- ditches gray square to have a cleaner look
- green button is now a gray with transparency

outmsgzqi6lrq

davibe:
@averissimo your work is great. So fare i have only looked at the super-amazing screencasts. This is what comes to my mind
- maybe "Drop here" overlay image should be limited to the conversation instead of covering the entyre application
- maybe the play (>) icon is not the best for "upload). I think something shaped like this would look better (besides color etc) http://www.freeiconspng.com/uploads/upload-icon-11.jpg

averissimo:
These are the icons available in the material icons font. The last one is the official hangouts icon for this purpose

image
image
image

Not really sure which is better, I kinda like the cloud upload as it has some density. Some more options here (all CreativeCommons):

https://thenounproject.com/search/?q=send

averissimo:
moved the overlay to the conversation. drag target is also just the conversation, i.e. cannot drop in contacts :)

averissimo:
Does not work with GIFs (as the current implemented workflow always converts images to PNG)

averissimo:
Electron's native-image does not support GIF yet https://github.com/electron/electron/issues/5078

It is possible to implement it by changing the clipboard module to a nodejs

Or we could just keep this as known bug and wait until electron implements it.

Bacto:
Hi @averissimo,

Do you know a nodejs module that can copy a GIF to clipboard?
I can't find one unfortunately :(

averissimo:
@Bacto I think electron webContents has a function that copies an image at position X,Y.. I haven't tried it it yet though so I don't know how well it works with static images or gifs

averissimo:
I have no idea about a general nodejs module

image

Bacto:
Unfortunately I want to copy an animated GIF so it will not do the job.
But thank you for your reply :)

averissimo:
so that functions copies the image as a NativeImage?

averissimo:
yeah I just tested it, it copies the frame it is on, never the entire gif :\

anyway, thanks for sparkling my interest in this, we'll probably add copy image feature soon

PR #418, opened 2016-10-15
Fix window position on startup

averissimo:
Having a blank window that shifts position and then changes to normal interface when starting YakYak has been bothering me for a while

This should fix it, both the positioning and the blank window in normal startups.

I've tested this PR with:
- clean userData (removing `.config/yakyak`)
- already logged in user

Can't think of any other scenario where it is relevant, help anyone?

Also tested it with a imposed bug in `menu.coffee` and it shows an empty window (which I think is a good thing as the user sees something that is nothing -- being able to report it)

ps. removed extra `;` as bonus :)

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -79,11 +79,7 @@ loadAppWindow = ->
     mainWindow.loadURL 'file://' + __dirname + '/ui/index.html'
     # Only show window when it has some content
     mainWindow.once 'ready-to-show', () ->
-        if !global.windowHideWhileCred? || global.windowHideWhileCred != true
-            unless process.platform is 'darwin'
-                mainWindow.setAutoHideMenuBar(true)
-            mainWindow.show()
-    # short hand
+        mainWindow.webContents.send 'ready-to-show'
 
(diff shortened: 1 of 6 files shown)


averissimo:
Added a startup screen (see end of this comment for screencast)

Startup screen does:

- Hide the dialog in the top right of the screen
- The "Connecting" text is always synced with that diaglog content
- The loading contacts appears when a successful connection is made (change happens on `update:connection`
- The contacts are considered loaded when `addentities` is finished
  - **if there is a better place to put this, please let me know**
- Nice fade out animation to show real interface

Preview:

*note that in the preview the animation shows the logo and text longer than it actually appears (this is a glitch from a low frame per seconds in gif)*

outgnmnkmdmv7

averissimo:
One last commit to help with debugging when UI does not render correctly.

Commit aa883a2 show mainWindow's menu, until event ready-to-show is triggered (which hides it)

In my environment if an error occurs the window is shown, in this case, the menu is still accessible and the user can open the DevTools

Issue #196, opened 2016-06-10
Missing user icons in 1.2.3


adamtheturtle:
https://github.com/yakyak/yakyak/releases/tag/v1.2.3 says "Fix missing user icons" but I still have the issue.

Thanks for the software 😄

davibe:
screenshot please ?

adamtheturtle:
![screenshot 2016-06-10 11 22 44](https://cloud.githubusercontent.com/assets/797801/15961678/dc311dde-2efd-11e6-9936-8d5a1738a07a.png)

OS X 10.11.5

davibe:
it does not happen with my contacts.
It would be interesting to know the url of that picture. It looks like an url has been set but it is invalid. cmd+alt+i fires up an inspector if you care to investigate a bit

adamtheturtle:
```
https://plus.google.com/_/focus/photos/private/AIbEiAIAAAAjCM2u98nbiIzp4QEQjZDF6aL5oeSnARiSs6WEjbHjkpEBMAGWIHIxPaIOAV7rgGpmL-LMbXHx3w Failed to load resource: the server responded with a status of 403 ()
```

![screenshot 2016-06-10 11 33 35](https://cloud.githubusercontent.com/assets/797801/15961868/3214f512-2eff-11e6-8216-d4263cd14fdf.png)

davibe:
which is a valid image. It failed to download somehow

<img src="https://plus.google.com/_/focus/photos/private/AIbEiAIAAAAjCM2u98nbiIzp4QEQjZDF6aL5oeSnARiSs6WEjbHjkpEBMAGWIHIxPaIOAV7rgGpmL-LMbXHx3w" />

davibe:
have you tried to logout and login again ? Or to close the app, delete completeliy `~/Library/Application Support/yakyak`, restart it

adamtheturtle:
@davibe Thank you for your help. Sorry it took me so long to respond - when I saw the message I wasn't in a position to log out and then I forgot about it. That didn't work, but deleting `~/Library/Application Support/yakyak` did work.

adamtheturtle:
Note that after trying logging out and before trying the deletion, I upgraded to 1.3.0 and the problem persisted.

jraoult:
FYI, I had the same problem and log out + removing `~/Library/Application Support/yakyak` did work for me.

sander-adhese:
[...]

PR #428, opened 2016-10-27
fallback to initials when thumbnail image not available

averissimo:
It was reported in issue #196 that some user's avatar do not show up with an error on the console log.

These were replaced by a generic image. **This fix uses the user's initials instead.**

This problem seem to only affect contacts that have a "private" avatar, that requires a login in google's infrastructure to get.
### TODO:
- [x] Messages show either all initials or all images
  - Problem is that when using the same logic as contact list / conversation add there is a bigger bug
- [x] Group chats have repetitive initials 
- [x] Details of the chat show default image
- [ ] Avatar has a lag changing when switching between conversations that have a valid avatar and problematic one

### DIFF
--- a/src/ui/css/yakyak/convadd.less
+++ b/src/ui/css/yakyak/convadd.less
@@ -1,4 +1,15 @@
 .convadd {
+  .initials {
+      border-radius: 50%;
+      width: 50px;
+      height: 50px;
+      text-align: center;
+      font-size: 17px;
+      line-height: 50px;
+      color: var(--white);
+      background: var(--grey);
+      text-transform: uppercase;
+  }
   padding: 20px 20px 0;
   h1{
     margin-top: 30px;
--- a/src/ui/css/yakyak/convlist.less
+++ b/src/ui/css/yakyak/convlist.less
@@ -1,4 +1,3 @@
-
 .convlist {
     .label {
         font-size: 9px;
--- a/src/ui/css/yakyak/messages.less
+++ b/src/ui/css/yakyak/messages.less
@@ -14,6 +14,14 @@
         text-align: center;
         color: var(--darkgrey);
     }
+    &.fallback-on .ugroup .sender {
+        img {
+            display: none;
+        }
+        .initials.fallback {
+            display:inherit;
+        }
+    }
     .ugroup {
         margin-top: 17px;
         position: relative;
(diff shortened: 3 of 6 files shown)


sander-adhese:
Does not work, I can see that `initials = initialsof entity[p.id]` generates the correct initials, but they aren't shown instead of the broken image.

averissimo:
It works on the conversation list? I was assuming the messages were using that one and did not test it.

Sorry about that, working on a global fix :)

### Without fix

selection_00129

### With fix

selection_00130

sander-adhese:
Hah yes it works on the conversation list, I didn't notice that at first. Well done :)

averissimo:
I think there is still a bug when having one of these users in a group chat

One of these should be VF

image

sander-adhese:
Still some issues, some photo's are overwritten with the initials (I can still see the right picture in the conversation list, not in the message window), some group conversations have multiple times the same initials in it, details of a chat is ok. Will try to think of other uses.

averissimo:
ahhh.. I have to think of a better solution then. Completely ignored the avatar that were supposed to exist.

The problem on implementing the same logic as in contact list is that for some reason when changing conversations initials were inherit from the previous conversation (only latest avatar showed correctly, others took a while to replace)

I'll update the first comment with a TODO of fixes.

Thanks a lot @sander-adhese

sander-adhese:
You're welcome!

averissimo:
I'm getting OCD over this xD

I'm not 100% happy with implementation, think I still need to clean it up, as it uses two approaches depending on messages or "conversation list"/"details".

### Fixed group chatting

by using a `data-initials` on the element, and setting onerror according to this `data-initials`

image

### Fixed messages

Although it seems to be a lag (microseconds) when switching between conversations that had a valid avatar and the ones that have a problematic avatar

image

averissimo:
@sander-adhese , can you test this when you have the opportunity? :)

Thanks

sander-adhese:
Sorry for the delay, yesterday was a holiday ;-) 

It works like a charm, thank you for your work! Well done 👍

sander-adhese:
Probably unrelated, but I can't cut and paste with my keyboard (apple keyboard, macOS Sierra) in the build of this branch, it still works by rightclicking though.

PR #429, opened 2016-10-27
[NEED HELP] Shows who has seen each message group

averissimo:
This is a big feature that was missing from YakYak and is implemented and working.. kinda of
### TODO:
- [ ] only works after the other party in the conversation has an action**, i.e. sends a message. (more details below)
- [x] Avatar of most recent message groups (not all though) disappear
- [ ] Decide on showing "seen" on each message group in conversation, or just the latest one (like hangouts)

All items require help, but I guess the most important one is the first, problem described below:
### Biggest problem:

For some reason `latest_read_timestamp` is initialized at `0` when opening the conversation (regardless of the history).. **I need help figuring this out, on what to use to get this value well initialized.**

note: it adds a seen avatar to each message group if that is seen (this can be customizable via an option to show all, show only last)

it has some quirks that need to be solved, I'm seeing the avatar sometimes disappearing and wiu

image

### DIFF
--- a/src/ui/css/yakyak/messages.less
+++ b/src/ui/css/yakyak/messages.less
@@ -19,6 +19,31 @@
         position: relative;
         width: 66%;
         margin-left: 25px;
+        .seen {
+          text-decoration: none;
+          font-size: 11px;
+          color: var(--darkgrey);
+          display: inline-block;
+          margin-right: 25px;
+          img, .initials {
+              object-fit: cover;
+              border-radius: 50%;
+              width: 25px;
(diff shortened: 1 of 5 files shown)


averissimo:
This feature is only active when `View -> Experimental -> Show seen status in messages` is activated

The biggest problem described above is from hangupsjs (and also hangups), so it may be a while since it is resolved.

I guess this pull request will be postponed until then, or available as `experimental`

Personally I will use it and I'm making it available via my personal branch or this Pull Request

PR #430, opened 2016-10-28
Fix on not fitting images in conversations list

aPoCoMiLogin:


aPoCoMiLogin:
it changes not fitting images:

before

to `object-fit: cover;`:

before

Issue #450, opened 2016-11-27
Sending Multiple Photos/Images


mattsigal:
I've been using YakYak for a while now (on OS X, 10.11.6), and come across an issue that I do not believe is open yet. 

Basically, when I use the "Attach Image" function, it works perfectly. The choose file dialogue appears, I select the image, and it uploads/attaches without issue.

However, if I switch conversations and try to send the same image to another friend, I click "Attach Image", the dialogue box appears, and I can select the image... but then nothing happens. It doesn't actually upload/send.  If I quit YakYak and restart the app, I can again send one image.

This issue has been present for me since at least 1.2.0, and I've recently replicated it on 1.3.1, 1.3.2, 1.3.3, and the latest Beta2.

averissimo:
great catch, simple fix in soon to be available beta3 (as soon as it uploads)

averissimo:
1.4.0 released

PR #454, opened 2016-11-29
Fixes problems reported in 1.4.0-beta1 and beta2

averissimo:
- Copy/Paste/Cut working on MacOSX
- Native about window problem solved, now calls on new "About" (MacOSX)
    -Help menu item is removed from MacOSX
- #450
- About window is now created on demand instead of being pre-loaded at the beginning.

(this is the code for 1.4.0-beta3 test release)

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -65,8 +65,8 @@ global.forceClose = false
 quit = ->
     global.forceClose = true
     # force all windows to close
-    aboutWindow.destroy()
-    mainWindow.destroy()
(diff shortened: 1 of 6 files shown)


averissimo:
@davibe sorry, forgot about this.

When hangupsjs is updated with pull request yakyak/hangupsjs#59

package.json should be updated to the actual release, instead of my fork of hangupsjs

davibe:
ideed that's one of the things i wanted to ask

Issue #76, opened 2015-10-26
Link message box to conversation


msphair:
If I type a message into the message box but don't send it and i switch to a different chat, the message stays there.

Can the contents of the message box be linked to a conversation?

kethinov:
+1

This is something I've been meaning to mention as well.

maxkueng:
+1

This is also the way the Hangouts app does it, IIRC.

algesten:
would be nice

xedef:
+1

averissimo:
#458 should implement this.

previous values are stored in `viewstate` object

I believe I tested with all options to switch conversation:

- clicking on a conversation
- selecting on menu
- keyboard shortcut

If i missed any please let me know :)

averissimo:
done! from 1.4.0

PR #458, opened 2016-11-30
clears input when switching conversations

averissimo:
it also remembers when returning to the conversation

#76


averissimo:
@davibe it was missing an important commit :\ I've pushed it to master

PR #462, opened 2016-11-30
Fix bad initials / avatars in convlist and convadd

averissimo:
This is mostly an implementation update.

user's avatars were being generated in 3 different methods, this unifies it by using `drawAvatar` from util.coffee


davibe:
conflicts!

averissimo:
fixed!

davibe:
no :P

averissimo:
it was a bad idea to correct indentation and break some code in multiple lines -_-

should be ok

PR #476, opened 2016-12-03
2 fixes on new conversations and removal

averissimo:
### New conversations via new message received

- When new conversations appear, they have no avatar on the conversation list and impossible to open details

Handle to add *client_conversation* was already available, but it was a bit conservative (only adding when conversation did not exist)

When a conversation starts via a new message received a conv. skeleton is created thus rendering the handle useless

### Removal of conversation

after confirming that we want to remove, changes to top conversation

previous behavior kept interface to remove user (that was already removed)

### DIFF
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -48,6 +48,8 @@ handle 'init', (init) ->
     require('./version').check()
 
 handle 'chat_message', (ev) ->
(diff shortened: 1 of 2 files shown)


averissimo:
had to make a correction to part of this commit as unread count was not correct.

pushed to master

see 7b97e167bc7b10de4e5a2f8e8a04c00c48f2c68a

Issue #471, opened 2016-12-02
Hangouts logo on splash screen changes size on star up


mosbror:
This is really a minor bug.
The Hangouts logotype on the splash screen changes size when switching between "Connecting" and "Loading contacts".

[yakyak_ssize.mov.zip](https://github.com/yakyak/yakyak/files/626603/yakyak_ssize.mov.zip)

averissimo:
it doesn't on linux

cool catch :)

averissimo:
corrected in pull request #472 

thanks for reporting 👍

averissimo:
@mosbror actually, can you test if the fix works?

Find the binary here:

https://github.com/averissimo/yakyak/releases/download/test-icon/yakyak-472-1.4.0-darwin-x64.zip

averissimo:
actually, that can be a result of having a different zoom level set in yakyak

can you confirm that happens if zoom is at 100% (ctrl+0) ?

Silveran:
It works well for me with your version, just a little drop of the logo margin-top (not its size),it seems that the "Loading contacts" wording changes its font-size during the splashscreen.

averissimo:
i can't fix this blindly, seems like in Mac OS X the relative fonts and margins are changed just after connecting.

Can't reproduce this is linux or windows 7

needs someone to look at this :)

Silveran:
It's very better than before with your patch, don't worry, it's not important ;)

mosbror:
I agree with Silverian.

averissimo:
That's the first thing new users actually see when they try YakYak, should be clean :)

**Can you try this:** https://github.com/averissimo/yakyak/releases/tag/sandbox

Silveran:
@averissimo It works, you rock !!

averissimo:
PR #477

Thanks for reporting and feedback, closing the issue

PR #477, opened 2016-12-04
"Loading contacts" text changed sizes and improvement on about

averissimo:
Former PR #472 did not fix the problem tracked in #471 completely

**Somehow the relative font-size changes after connecting in Mac OS X only.**

This is reaaaaaaaally strange, but was corrected by using absolute dimensions for font-size and margins.

-----

plus, changing the about window to a frameless one that closes when clicking on top right corner or outside the window

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -379,12 +379,17 @@ app.on 'ready', ->
           show: false
           parent: mainWindow
           resizable: false
+          frame: false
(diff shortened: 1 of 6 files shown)


davibe:
conflicts again

averissimo:
solved the conflict

btw, this is how it looks

image

Issue #439, opened 2016-11-16
Unable to switch languages with alt+shift


browseman:
Linux Debian7 Xfce4.8
There is a problem when switching language with alt+shift combination. When the switch combination is pressed, the hidden menu is opened and the language is switched, but it's unable to continue writing unless you hide back the menu (which usually is opened with alt key).
Is there a way to have an option to disable this behavior of the hidden menu? Hope my explanation are sufficient.

averissimo:
what's the full key combination? alt + shift + **??**

browseman:
Only "alt+shift", there is no additional keys involved.

averissimo:
does clicking on the three lines 'menu' work on setting it while yakyak keeps running?

this option is not permanent, but if this works maybe it could become permanent :)

browseman:
I don't think I understand you correctly. But when I click on "three lines" button the menu is falling down. Then I was able to choose from it until I press some button of the keyboard. The "alt+shift" combinations does not simply open the menu but is like actually

averissimo:
@browseman which version of yakyak are you using? my comments below are for 1.4.0-beta3

I tried installing xfce and can't reproduce this in my laptop with Xfce 4.12.3 (current version of xubuntu-desktop as of 17.04)

The point about the icon is that once I click on the "three stacked lines icon" the menu either permanently appears or disappears *(until clicked again)*. I believe this would be the same as having an option to always show the menu

Alt produces nothing (which I think is strange, as it should highlight the menu)

Alt + Shift produces nothing and doesn't switch keyboard languages, how do you go on setting this?

browseman:
@averissimo - I'm using v1.2.5 (according to the version file in the prebuilt binaries that I've download).

My version of Xfce is a bit older version 4.8 as already mentioned in my first post, but I doubt that is the reason. Most probably the reason is in the older version of the app. I will try to test with the latest one.

This behavior with the "three stacked lines icon" does not work like you describe it with v1.2.5. But with pressing the:

- Alt key - only opens the menu
- Alt + Shift keys - opens the menu with first dropdown also opened.

I will give a feedback after test with the new/latest version.

browseman:
I've tried with the latest version (git clone https://github.com/yakyak/yakyak.git). Currently the behavior of the "three stacked lines icon" is working as it should. The menu is constantly stay on or off regardless of the pressing of other buttons. The behavior of:

- Alt key - is changed, currently the single pressed alt key doesn't do any thing.
[...]

PR #479, opened 2016-12-04
#439, alt+shift issue workaround

iovd:
Empty menu entry with '' label added to prevent menu opening while alt+shift layout switch.

This is NOT 'production ready' fix lol, but it can help quickly some users with same trouble.


averissimo:
@davibe can you test this on Mac OS X?

I've been doing so in Gnome Shell for a while and it shows no problem

src/ui/views/menu.coffee
```diff
@@ -432,6 +432,10 @@ templateWindow = (viewstate) -> [
 templateMenu = (viewstate) ->
     [
         {
+            label: ''
+        }
+        {
```
averissimo: can you make this a linux only fix?

I'd recommend that you create a variable at the top 

isLinux = platform == 'linux'

and then add a `if isLinux` after the '}' of that label

averissimo: additionally

could you add as a comment the original issue number (#556) as to trace back the change in the code

looking at it without context it could look like a strange menu to have :)

averissimo:
And sorry for leaving this hanging for so long, I've been testing for more than a month and noticed no problem in Gnome shell.

davibe:
i trust this is useful on Linux :P

PR #486, opened 2016-12-05
Corrections on feature 'seen'

averissimo:
### Correction from new feature

**Bug:** Show only last seen avatar is not working properly once you start to scroll up and history is fetched. Duplicate avatars are shown and sometimes the last one is not shown.

**Solution:** Show all seen avatars throughout the conversation

Also makes feature more robust, everytime the conversation gets focus it also requests an update on the metadata and updates it, instead of relying on getting it from history

### General correction in YakYak

**Bug:** I've been noticing through my usage of YakYak that sometimes YakYak starts scrolling up alone without input.

I found out why today why while implementing a more predicted way of getting the read_state: **We are using duplicate ids in the interface.**

 The message id is used in message pane, but also in the conversation list on the left. And this id is important as it is used when redrawing the conversation (to refresh it and get back at the same point). 

**Solution:** Message snippet id now has a prefix

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -298,6 +298,10 @@ app.on 'ready', ->
     # no retries, dedupe on conv_id
     ipc.on 'setfocus', seqreq (ev, conv_id) ->
         client.setfocus conv_id
+        client.getconversation conv_id, new Date(), 1, true
+        .then (r) ->
+            ipcsend 'getconversationmetadata:response', r
+
     , false, (ev, conv_id) -> conv_id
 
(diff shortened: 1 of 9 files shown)


davibe:
awesome

averissimo:
It's working without problems so far, change in id didn't seem to break anything :)

Message snippet is also updating with conversation

averissimo:
@davibe been using yakyak a lot today since the last commit was pushed and nothing seems broken :)

should be ready to be merged and tested in a pre-release :)

PR #487, opened 2016-12-05
use runSequence to compile perfore starting to deploy

averissimo:
calling on `npm run deploy:linux-x64` without compiling first would create a broken binary

using runSequence allows to run sequentially and allow for the default target to finish before starting to package.

thanks to @nerijus for picking this up :)

ps. @davibe I'm going to merge this now


averissimo:
@davibe btw, this is the cause that the linux-x64 binaries don't have node_modules!!

PR #499, opened 2016-12-08
Add wine warning when building windows binaries

averissimo:
When building in Linux and Mac OS X it checks if it has wine installed and print a warning to install

If wine is installed it gives a notice

image

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -315,7 +315,8 @@ deploy = (platform, arch) ->
             if ev?
                 console.log ' IMPORTANT for building Windows binaries outside Windows'
                 console.log ''
-                console.log ' It is required to have wine installed and configured. If problem persist try a different wine prefix.'
+                console.log ' It is required to have wine installed and configured.'
+                console.log ' If problem persist try a different wine prefix.
                 console.log ' -------------------------------------------------------'
 
                 console.log ''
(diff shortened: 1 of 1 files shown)


nerijus:
I am not sure if 32 bits specifically is needed, as I was able to use 64 bit prefix wine. But if rcedit.exe is 32 bit app, then technically yes, wine which is able to run 32 bit apps is needed.

averissimo:
**can you build the 32bit binary with 64bit wine?**

do you think 32bit reference should be removed and just say wine?

or rephrase to something like:

```
... please be sure to have wine installed and configured.

(if wine still fails, try using both wine 32 and 64 bits architecture)
```

nerijus:
Yes, just wine is OK.
"(if wine still fails, try using both wine 32 and 64 bits architecture)"
No, for me 64 bit prefix helped because my 32 prefix was wrecked. So just wine is OK, or something like "try with a new/empty prefix".

nerijus:
> can you build the 32bit binary with 64bit wine?
I built 32 bit binary with 64 bit wine prefix, but I am not sure if wine exe 32 or 64 bit was used. But prefix was 64 bit.

averissimo:
updated

nerijus:
"If problem persist try a different wine prefix" should be "If problem persists try a different wine prefix".

davibe:
Is it yakyak job to detect wine absence and give warnings ? shouldn't it be a job of electron-packager or one of its dependencies ?

averissimo:
true, I'll open a issue there instead

averissimo:
https://github.com/electron-userland/electron-packager/issues/357 open issue and currently being discussed

PR #500, opened 2016-12-08
npm start | npm run dev | no_zip packaging | electron-prebuilt deprecated

averissimo:
- electron-prebuilt package is deprecated 
- checks if NO_ZIP environment variable is set to skip zip creation in packaging
- adds `npm start` (no run in the middle, compiles and starts) and `npm run dev` targets (inspired by electron demo: https://github.com/electron/electron-api-demos )

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -306,6 +306,8 @@ deploy = (platform, arch) ->
         if err?
             console.log ('Error: ' + err) if err?
         else if appPaths?.length > 0
+            if process.env.NO_ZIP
+                return deferred.resolve()
             json = JSON.parse(fs.readFileSync('./package.json'))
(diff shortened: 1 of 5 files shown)


averissimo:
the lint tab inside the devtron tab when running `npm run dev` has some nice tips on stuff we should eventually do

marked in red:

image

averissimo:
Added methods to catch these events in https://github.com/yakyak/yakyak/pull/500/commits/553c76e3aaaf5c7abbc8efbad602aa446bd1a5dd

averissimo:
@davibe the .travis file has as requirements node_js `0.12`

Ca we update this requirement to v6 or v4 I believe the travis check failed on this one because of this

davibe:
sure..

PR #501, opened 2016-12-08
Avatar color fixes

averissimo:
Add color yakyakgreen that can be used in css (for now i'ts used in the connecting sequence)

For large conversation, instead of just a number, added a '+' before

image

Avatar initials have a consistent look in different color schemes, bright color background, white lettering.

image

image

image

image

### DIFF
--- a/src/ui/views/convlist.coffee
+++ b/src/ui/views/convlist.coffee
@@ -33,7 +33,7 @@ module.exports = view (models) ->
                             drawAvatar(p.id, viewstate, entity)
 
                         if ents.length>4
-                            div class:'moreuser', ents.length
+                            div class:'moreuser', "+#{ents.length}"
                         if ur > 0 and not conv.isQuiet(c)
                             lbl = if ur >= conv.MAX_UNREAD then "#{conv.MAX_UNREAD}+" else ur + ''
                             span class:'unreadcount', lbl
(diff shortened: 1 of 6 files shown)


davibe:
why you want to force white color for initials and not use the theme color ?

averissimo:
Two reason

- The contrast was not good
- Visual consistency across all hangouts-related apps

But at the time I didn't weighted in the break with previous interface of yakyak itself, or dismissed as I personally think it looks better :)

I will revert this change as soon as I click on comment ;)

src/ui/views/convlist.coffee
```diff
@@ -27,13 +27,20 @@ module.exports = view (models) ->
...
+                            #   then draw them all, otherwise, draw 3 avatars
+                            #   and then add a +X , where X is the remaining
+                            #   number of people
+                            if index < 3 ||  ents.length == 4
+                                entity.needEntity(p.id)
+                                drawAvatar(p.id, viewstate, entity)
+                            else
+                                additional.push nameof entity[p.id]
+                        if ents.length > 4
+                            div class:'moreuser', "+#{ents.length - 3}"
```
averissimo: changed this to +X instead of showing the total number of persons in a conversation

**rationale:** I was interpreting this as the people in addition to the 3 avatars already drawn. Only by looking at the code I realised it was the total number of people in the conversation

I can revert this to previous behavior if someone doesn't agree with this :)

davibe:
I was just asking. I think it may be possible that a theme chooses to have dark text because it's using a light background for it.

PR #504, opened 2016-12-10
About window was calling all scripts from main windows needesly

averissimo:
Also catches exceptions from about window and prints to main process

### DIFF
--- a/src/ui/version.coffee
+++ b/src/ui/version.coffee
@@ -15,7 +15,7 @@ check = (noAlert = false)->
         return console.log err if err
         body = JSON.parse body
         tag = body.tag_name
-        releasedVersion = tag.substr(1) # remove first "v" char
+        releasedVersion = tag?.substr(1) # remove first "v" char
         localVersion = require('electron').remote.require('electron').app.getVersion()
         versionAdvertised = window.localStorage.versionAdvertised or null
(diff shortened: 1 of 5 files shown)


src/ui/version.coffee
```diff
@@ -15,7 +15,7 @@ check = (noAlert = false)->
         return console.log err if err
         body = JSON.parse body
         tag = body.tag_name
-        releasedVersion = tag.substr(1) # remove first "v" char
```
davibe: @averissimo why ?

averissimo: getting error on this line, replaced by following one

image

davibe: i know but it should not fail unless we have done something wrong with the releases

averissimo: the about window is not implemented properly in the trifl framework as it is a static window.

I didn't know better at the time and didn't implement separated from the app (in the sense that as soon as the app was being drawn the first time, it also was drawing the about.. in turn was requiring version check that would fail.

By some reason the request would not stop in the `if error` condition and would throw an exception at this point. As there is a `releasedVersion?` check next that would catch this gracefully.

Anyway, I need to talk to you about how best to implement a static window in trifl and probably refactor the about window

PR #509, opened 2016-12-13
subtle links to the install hints

l00mi:


davibe:
I'd rather link to the binaries instead of the wiki explaining how to build. It's what we did in the past but since it's boring to always update links i stopped doing it.

l00mi:
My intend is to make it visible how you can install YakYak with Homebrew on macOs which is a really convenient way to stay up-to-date.

davibe:
It's the subtle part i think i don't like. It has to be clear. Does this look acceptable to you ?

l00mi:
Perfect! cheers

PR #513, opened 2016-12-14
improves debug capabilities to show hangups debugging

averissimo:
Allows to have a more verbose debug when running with "--debug flag"

* `hangupjs` debugging appears on the terminal. This alone can reveal some hidden problems (see first screenshot for a current master branch run)
* allows to have some hidden messages easily available, such as:
  * "attachment" that was appearing in the console
  * __redraw() call -- 44 times just for initial boot-up__ (see screenshot 2)

### Screenshot 1

image

### Screenshot 2

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   "homepage": "https://github.com/yakyak/yakyak",
   "dependencies": {
     "autosize": "^3.0.8",
+    "bog": "^1.0.2",
     "hangupsjs": "^1.3.4",
     "mime-types": "^2.1.12",
     "moment": "^2.10.3",
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "homepage": "https://github.com/yakyak/yakyak",
   "dependencies": {
     "autosize": "^3.0.8",
-    "bog": "^1.0.2",
+    "electron-is-dev": "^0.1.2",
     "hangupsjs": "^1.3.4",
     "mime-types": "^2.1.12",
     "moment": "^2.10.3",
@@ -52,6 +52,7 @@
     "chai": "^3.2.0",
     "coffee-script": "^1.9.2",
     "devtron": "^1.4.0",
+    "bog": "^1.0.2",
     "electron-packager": "^8.3.0",
     "electron": "^1.4.10",
     "gulp": "^3.9.0",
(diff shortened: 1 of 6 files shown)


averissimo:
__I have an open question, should the debugging capabilities be included in the binaries?__

The only upside I see is in `main.coffee` to debug hangupsjs (it would be a lot easier to understand when people say that I can't login in yakyak)

In the main window, I'm kinda neutral on having or not, it could be useful if we anticipated all the necessary information on all the problems.. and that's not going to happen LoL, at least I don't have that power in me.

davibe:
With trifl we have a virtualdom in place which diffs against the real dom and only applies relevant changes to it. Because of this 44 redraws should not be a problem. Sure there is some room for improvements but i don't think that causes anything noticeable to the user, expecially considering it's just at boot.

Regarding logs I don't like too much the idea of dispatching logs from one process to the other. In another PR you have already brought some of the renderer logs to the background process. That was useful for a specific reason being: if a window crashes or is hidden i can't debug it with the inspector.
Now you're trying to do the same in the opposite direction. creating inter-process communication for every log can be itself cause of poor performances or strange issues (i.e. you may have a bug triggering endless logs in the background process which would cause the renderer process to hang. How do you know which one to debug?). Even if we want to do this i would avoid enabling it in production.

averissimo:
**Redraws**

it could be improved as you said, but since your last commit detecting the animation as the main source of cpu usage i guess this is really low impact  

Part of my motivation bring this up was that at some times my yakyak was blocked for a couple of seconds.. especially after using it for long periods. I had suspicions that a weird timing of a redraw (or excessive) could be a cause.

**Inter-process logging**

You make an excellent case! In case we want to follow through with logging other libraries, I believe they should be really strong exceptions and not the norm.

`hangupsjs` is the obvious core library, which in debug mode captures all the communications and does it regardless of our implementation of yakyak. It just does not appear in the terminal.

Considering the number of "I can't log in yakyak" / "permanent sign in loop" issues we have open, it could help to understand those as they are hard to replicate.

Still the extra burden on **ALL users** is difficult to justify and could create unnecessary problems that I can't anticipate.

I guess we can keep the module only in development, and only load `bog` module if installed (much as `devtron` is implemented).. this would be flexible enough to be used in dev or for users to customize their installation (it has no dependencies, so it's easy to just plug-in)

_ps. Finding out that it was able to log hangupsjs was a suprise, lol.. I was testing diferent logger modules and when activated this one on the main process 'puff' that stuff appeared._

Issue #518, opened 2016-12-20
Failed to execute 'appendChild' on 'Node' -- Problem starting 1.4.1 on Win8.1


remizyaka:
When trying to start app see such error in command line on Windows 8.1.
```
Error on YakYak window:   
 { columnNo: 5,
  error: { code: 'MODULE_NOT_FOUND' },
  lineNo: 457,
  msg: 'Uncaught Error: Cannot find module \'autosize\'',
  url: 'module.js' }
--- End of error message in YakYak window.
```
Program window won't appear but I see yakyak.exe process running

davibe:
did you download binaries or have you built it yourself

remizyaka:
downloaded binaries, tried both x64 and ia32

WiXZlo:
I have some error on my Fedora 25. I download pre-built binary.

averissimo:
Confirmed this in Win32 and all the binaries seem to have this problem

Don't know why but there's a folder missing in the `autosize` module

`resources/app/node_modules/autosize/dist`

The win32 binaries have been re-uploaded and are working correctly in my Windows 7 VM (32bits)

@remizyaka can you confirm if the win32 binary are working correctly? _(other binaries are being uploaded)_

davibe:
@averissimo This comes from the folter of the machine i used to buid packages
```
ubuntu@Ubuntu-1404-trusty-64-minimal:~/yakyak/dist$ find . -name autosize
./yakyak-linux-x64/resources/app/node_modules/autosize
./YakYak-darwin-x64/YakYak.app/Contents/Resources/app/node_modules/autosize
./YakYak-win32-ia32/resources/app/node_modules/autosize
./YakYak-win32-x64/resources/app/node_modules/autosize
./yakyak-linux-ia32/resources/app/node_modules/autosize
```

this is the zipfile that I uploaded after being unzipped 
```
ubuntu@Ubuntu-1404-trusty-64-minimal:~$ find YakYak-win32-x64/ -name autosize
YakYak-win32-x64/resources/app/node_modules/autosize
ubuntu@Ubuntu-1404-trusty-64-minimal:~$ ls `find YakYak-win32-x64/ -name autosize`
LICENSE.md  changelog.md  package.json  readme.md  src
ubuntu@Ubuntu-1404-trusty-64-minimal:~$
```

davibe:
@averissimo on the machine i used to build `yakyak/dist/YakYak-win32-x64` is missing the folder you mention indeed

[...]

PR #521, opened 2016-12-21
nodejs version bump and show error window on pre-startup error

averissimo:
Changes:

- removes console logging of attachment in messages
- our current version of electron requires v6, so update .travis.yml
- users should always see something when an error appears

For the last change, the error window is static and defined explicitly in the `index.html`, which is replaced as soon as the app is ready to show.

The message is only shown if an exception is caught before the event `read-to-show`

How to test:

just delete app/node_modules/autosize/dist :) _(see #518)_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
 node_js:
-  - "0.12"
+  - "6"
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


averissimo:
This is what appears when an error occurs loading the app.

This is embedded in the `index.html`, which is completely replaced if there's no problem in the main process or with modules.

image

averissimo:
to test this, just delete `app/node_modules/autosize/dist` :)

first link opens inspector, second our issue's tracker in the default browser.

PR #522, opened 2016-12-22
Fix trifl

averissimo:
This is a big one :)

### Startup is contained in 7a118cf and 94eb199

Created a new region in `applayout` called `last` where startup is displayed, so that it covers the whole window.

startup element is removed once the fade out transition ends (by listening to `transitionend` event and calling `action 'remove_startup'`)

### About in 6ebaffe

I've created a new state called STATE_ABOUT that will render the about window in the right panel. And that's pretty much it.


averissimo:
Example for about pane

image

PR #530, opened 2017-01-04
bring back Noto font

averissimo:
partialy reverts: 6245097febf847cd9d027c04df5000de3ae3736b
works in linux


davibe:
on osx i don't see any difference

davibe:
i merge this and hope for the best

PR #533, opened 2017-01-04
Travis CI builds binaries and creates github releases automatically

averissimo:
- Deploy is only triggered with new tags
- Creates pre-releases by default and not official ones

It also overwrites previous files, so there's no change for the pre-release workflow

__REQUIRES:__

- API_KEY environment variable set in travis-ci.org with github API key
    - can be obtained using travis command line `travis setup releases`
    - important: the key must be unencrypted

Example:

image


averissimo:
I've added the branch to yakyak/yakyak and configured travis with my github key

__The automatic deploy can be tested by tagging something in that branch and push it__

It will create a pre-release tag or update if we are force tag and pushing it

davibe:
this is <3

averissimo:
oops it was deleted automatically after I removed the branch from my fork

restored here: https://github.com/yakyak/yakyak/pull/546

PR #545, opened 2017-01-09
Fix pathname encoding, #544

### LINKED ISSUE #544: App can't run from a path that has `#` in it
e.g. `c:\electron\#net\yakyak`

refack:
Fix #544

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -7,6 +7,10 @@ path      = require 'path'
 tmp       = require 'tmp'
 session   = require('electron').session
 
+
+[drive, path_parts...] = path.normalize(__dirname).split(path.sep)
+global.YAKYAK_ROOT_DIR = [drive, path_parts.map(encodeURIComponent)...].join('/')
+
 # test if flag debug is preset (other flags can be used via package args
 #  but requres node v6)
 debug = process.argv.includes '--debug'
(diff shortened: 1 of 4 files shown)


src/main.coffee
```diff
@@ -114,7 +118,7 @@ app.on 'ready', ->
         height: 590
         "min-width": 620
         "min-height": 420
-        icon: path.join __dirname, 'icons', 'icon@32.png'
+        icon: path.join YAKYAK_ROOT_DIR, 'icons', 'icon@32.png'
```
davibe: are you sure about this ? is icon a plain path or an url ? I believe a plain path should not be encoded.

refack: Investigating...

refack:
Yep, you were right, too much encoding.
Now it works, but I'm not sure about the `toast` notification (they stopped working either way)

davibe:
Maybe we manage `toast` notifications in a separate issue/pr ?

averissimo:
global options can be retrieved on the client side, i think `force_close` is one of them

refack:
I've seen it, but thought I killed it (client-side `undefined`). 1st try definatly fixed it, and the 2nd try probably regressed it back in.
I think the `toast` issue is related to the custom sound. Will spin off a new issue.

refack:
Can I get a hint as to what's the entrypoint to the client process?

davibe:
in main.coffee we create a BrowserWindow. We actually create one for login and then one for the main window. Anyway, the mainWindow will load `ui/index.html` which in turn loads the compiled `ui/app.coffee` so this last one is the entry point you want, i think. In that file we already use `remote.getGlobal`. Everything that lives in `ui/` is about the renderer process.

PR #548, opened 2017-01-10
move rendering of menu and dock out of redraw

averissimo:
Menu is rendering too many times if placed in `redraw()` function

This can be replicated by opening the menu and leaving it open a few seconds

This correction only renders menu when `viewstate` is updated as it originally was

### DIFF
--- a/src/ui/views/controller.coffee
+++ b/src/ui/views/controller.coffee
@@ -81,6 +81,7 @@ handle 'update:viewstate', ->
 
         menu viewstate
         dockicon viewstate
+        trayicon models
 
     else if viewstate.state == viewstate.STATE_ABOUT
         redraw()
@@ -124,7 +125,11 @@ redraw = ->
     input models
     convadd models
     startup models
-    trayicon models
+
+
+handle 'update:language', ->
+    menu viewstate
+    redraw()
 
 handle 'update:switchConv', ->
     messages.scrollToBottom()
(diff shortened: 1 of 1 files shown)


davibe:
what is the drawback of rendering "too many times" ?

davibe:
I propose this https://github.com/yakyak/yakyak/pull/549 what do you think

averissimo:
It is not the amount of redraws, since the animation _fiasco_ that was using too much cpu that I don't really care about how many redraws it take.

__The problem is__ that the menu, dock  and tray are being rendered with everything that is happening in the app.

__Example:__ when user has menu open and receives a message in YakYak

outxq7lo6ybpp

averissimo:
anyway, this pull request just returns those 3 lines as it was until 9cf83419a8d7d45a95c3da048f3bf5046d667768 (9 days ago)

where I accidentally removed them and did a mistake when reverting it (wrongfully placed the lines in redraw)

davibe:
Ok so the issue is:
When application menu is open and the app receives some event causing a redraw -> the app menu closes unexpectedly

This pull request restores some lines (which is good) but does not currently solve this problem, right ? Because during `viewstate.STATE_NORMAL` menu redraw happens anyway.

Maybe we need a way to skip menu redraws while menu is open. Maybe we should tell electron guys about this as i don't think it was happening before.

averissimo:
it was happening since I remember, before I was noticing it when trying to open the menu while YakYak was connecting.

While the app is running it is less noticeable as `viewstate` is not updated that often.

Issue in electron: https://github.com/electron/electron/issues/8385

PR #555, opened 2017-01-15
normalize coffee a smidge (pre decaf)

refack:
These tiny changes/improvements allow using `[decaffeinate](https://github.com/decaffeinate/decaffeinate)` to get pretty-pretty ES6 (node 4 compliant) output

### DIFF
--- a/src/ui/models/connection.coffee
+++ b/src/ui/models/connection.coffee
@@ -14,8 +14,6 @@ EVENT_STATE =
 TIME_SOME = 40 * 1000      # 40 secs
 TIME_ALL  = 10 * 60 * 1000 # 10 mins
 
-merge   = (t, os...) -> t[k] = v for k,v of o when v not in [null, undefined] for o in os; t
-
 info =
     connecting:     'Connecting…'
     connected:      'Connected'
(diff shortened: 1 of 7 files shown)


davibe:
why do you decaffeinate yakyak ?

refack:
I'm enjoying playing with compilers, but I'd take the challenge, and make your CoffeeScript compile as is 😉

refack:
[OT] why do you use the two proc approach (dispatcher & client)?

davibe:
I was not against the PR. Just wanted to undestand.

It's electron that separates those two concepts. I think at first we did it so that everything could survive even without any window open. Later we discovered we could just hide the window.

 I am not a big fan of this separation either and i think we could try to move yakyak from the 'server' process to the 'renderer' one. I am not sure what would break though. Maybe the renderer process does not have 100% of node-apis. I don't remember exactly

refack:
Got it (RE: separation)
As for the PR, fixing `decaffeinate` to deal with real-world-in-the-wild code is better after all (plus `hangupsjs` is way more challenging 😜)

PR #567, opened 2017-01-30
updates tray when unreadCount is called

averissimo:
also corrects some indentations


smkent:
I tested this briefly and it seems to work.

Issue #167, opened 2016-04-11
Yakyak not reporting current user as available on other clients


randymorris:
Using Yakyak Version 0.37.4 (1.2.0) I am getting complaints from coworkers that I am not showing up as available on official Google clients even though I am logged in and chatting with other people.

I don't believe this is specific to 1.2.0 as this has been mentioned once or twice in the past.  I had ignored it then thinking it was just an issue with their client lagging in presence updates, but the last few times have been after I had been signed in via Yakyak for hours.

Can anyone else reproduce?

vgavrila:
I can recreate. But is this a bug or an enhancement?

randymorris:
IMHO it's a bug.  Until there is a way to set your status I think you should be set to available.  As it stands, I have to run two different clients just to let people know I'm online.

davibe:
It's a bug on user's perspective. But since the code never supported this it's an enhancement for the project.

anevilsithlord:
Are there plans to include User Status in a future release?

arnriu:
@anevilsithlord yes definitely. It's on my list unless someone does it before ;)

Stanko:
I can recreate as well.
This would be really nice to get into the next release.

davibe:
so, do it and send us a pull request!

Stanko:
@davibe I actually already took a look, but I'm not sure where is the problem. I would appreciate advice where to start.

etrulls:
Is this still the expected behaviour? I keep getting messages from coworkers that I'm showing as offline.

NvrBst:
Also seeing this. The YakYak client is setting myself as "Invisible", so no one can see I am online. I'm using the latest YakYak Client (1.4.1). It seems to me that the client should show as "Available" or "Online" by default...

Everyone using YakYak is fine with never showing as online to anyone? This seems like a pretty big bug to me... I'm not too sure about anyone else, but assuming people are using only YakYak, not having anyone able to see if you're online ever (and thus probably not sending you messages)... then what's the point in having the messaging client up then in the first place?

If you want "Invisible" that's fine... but it should definitely not be the default behavior...

davibe:
I see myself as logged in (green) when using desktop hangout website from another user.

<img width="192" alt="screenshot 2017-01-12 18 15 06" src="https://cloud.githubusercontent.com/assets/123929/21908292/6de5687e-d913-11e6-98b3-1b81d8066a90.png">

NvrBst:
[...]

PR #571, opened 2017-01-31
setpresence with each connection cycle #167

averissimo:
Sets presence every time there is a connection event (and is connected)

#167

### DIFF
--- a/src/ui/views/controller.coffee
+++ b/src/ui/views/controller.coffee
@@ -15,10 +15,9 @@ handle 'update:connection', do ->
         # draw view
         conninfo connection
 
-        console.log 'Updating connection'
-
         # place in layout
         if connection.state == connection.CONNECTED
+            later -> action 'lastActivity'
             el?.hide?()
             el = null
         else if viewstate.state != viewstate.STATE_STARTUP
--- a/src/ui/views/controller.coffee
+++ b/src/ui/views/controller.coffee
@@ -15,6 +15,8 @@ handle 'update:connection', do ->
         # draw view
         conninfo connection
 
+        console.log 'Updating connection'
+
         # place in layout
         if connection.state == connection.CONNECTED
             el?.hide?()


davibe:
does it make sense to do it periodically ? like every minute if you're connected.

averissimo:
It takes around 30mins timeout, so every minute might be overkill. Attaching this to the handle `update:connection` seemed logical, but so does doing it periodically, I don't really favor one or the other.

You know the inner ways of YakYak better than me and whichever way seems better to you I can implement it :)

davibe:
lets start with this.

NvrBst:
Heya!

I'm still seeing this issue. I have upgraded to the latest pre-release (v1.4.2 - 9ade528) released 1 day ago.  I confirmed the change you made to set the presence on every _"update:connection"_ is in the code on my system:
```
ui/source/controller.js
      if (connection.state === connection.CONNECTED) {
        later(function() {
          return action('lastActivity');
        });
```
I deleted all the other YakYak's on my system, and ran this one. However, when I leave YakYak in the "Closed to Tray" state, my online status is set to Offline after a while. I have timed how long this takes, by moving my mouse around the YakYak client, then closing it to tray and seeing when it sets me to offline. It does it in exactly 15 minutes and 10 seconds. You have to log in and out of hangouts web interface to have the client refresh the online status of the list quicker. Else the web client doesn't update the people's status as often unless you relog in.

How often does this _"update:connection"_ happen in a _"connection.CONNECTED"_ state? Can the connection state be changed from "CONNECTED" to something other after a while? i.e. "IDLE" or "CLOSED" if no activity has been detected for awhile? Then it stops setting the presence? Maybe it doesn't go into this state every 15 mins?

Maybe setting _lastActivity_ periodically may be better? Every 10-minutes to stay within the 15-minute window?

averissimo:
The update:connection should be performed every minute or so, but I'm testing it right now to make sure. In 20 mins I'll send a new screenshot

image

What I'm seening is a huge difference on who is online between YakYak and gmail hangouts interface. That should be a new bug :(

averissimo:
The timer in `connection.coffee` is not being run with the hidden window keeping the connection active.

Just confirmed that with the inspector window open it works. I probably had that open during the tests :\

NvrBst:
Yea, running 1.4.2, I see that YakYak is getting confused on who is online and not. It says almost everyone is online. So some bug there in the new version. :)

Ah good to know you found the issue with the _presence_. :) Need to find a different place to update _presence_ then it sounds like. :P

averissimo:
I think associating it to the noop event from hangouts might be the way to go.

Why? it is a keep alive signal, so doing it after receiving is obvious

Problem is that it does keep the status as available... need to work on this more

averissimo:
let's move the conversation to #596 :)

PR #572, opened 2017-01-31
makes emojicategories readable by separating with newlines

averissimo:
While fixing something with emojis I stumbled with this one liner file, which makes it hard to edit/find stuff in atom

had to convert to this to look for what I wanted

__In any doubt, don't merge this__


davibe:
haha, so i don't merge.

Issue #573, opened 2017-02-02
Typing a URL right before a close parenthesis results in the parenthesis incorrectly being added to the link target


smkent:
Frequently I type messages in this format:

> Check out this thing (at http://example.com/)

Google Hangouts doesn't append the trailing ) to the URL:

![hangouts](https://cloud.githubusercontent.com/assets/2831985/22538460/56b89d10-e8c7-11e6-9759-0d4ff327429f.png)

yakyak does, which usually results in the link target being broken (since it's different than intended):

![yakyak](https://cloud.githubusercontent.com/assets/2831985/22538474/688a964c-e8c7-11e6-8bba-260125109837.png)

The second and third lines are incorrect; for example the Test 2 link target is `http://example.com/)` instead of the intended `http://example.com/`

yakyak's URL detection in composed messages should be updated to exclude trailing parentheses immediately following a URL.

(As a sidenote, Github's markdown renderer does the correct thing with these types of URLs -- see the example message near the top.)

smkent:
I looked slightly deeper. yakyak seems to pass of URL parsing to [gajus/url-regexp](https://github.com/gajus/url-regexp), which exhibits the same behavior. The URL parsing regex just consumes characters until a whitespace character is found: https://github.com/gajus/url-regexp/blob/master/src/url-regexp.js#L56

I'm not sure what the correct solution is here. url-regexp hasn't been updated since 2015, is the project still active? Are there any other URL parsing libraries that yakyak should consider using?

averissimo:
https://github.com/kevva/url-regex seems to do the trick

averissimo:
@smkent can you see if the implementation is correct?

I also added support for links in messages without protocol. Example: `www.github.com`

smkent:
This is definitely an improvement but it's not handling all the cases correctly. I wrote a more detailed comment on #573.

PR #578, opened 2017-02-06
Update url regex expression #573

averissimo:
allowing links in parenthesis and also open recognized links without a protocol `www.google.com`

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "request": "^2.67.0",
     "tmp": "0.0.28",
     "trifl": "0.0.10",
-    "url-regex": "^4.0.0"
+    "url-regex": "averissimo/url-regex#v5.0.0"
   },
   "devDependencies": {
     "chai": "^3.2.0",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "request": "^2.67.0",
     "tmp": "0.0.28",
     "trifl": "0.0.10",
-    "url-regexp": "^1.0.2"
+    "url-regex": "^4.0.0"
   },
   "devDependencies": {
     "chai": "^3.2.0",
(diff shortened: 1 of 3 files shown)


smkent:
Links without a protocol (ex. `www.github.com`) and the test case you listed in #573 with a URL in parentheses (`ada (http://www.google.com)`) seems to work. However, links in parentheses with a URI component still are linked incorrectly, e.g. "([http://www.example.com/bogus-url/)](http://www.example.com/bogus-url/\))"

davibe:
so i wait before merging

averissimo:
Upstream issue: https://github.com/kevva/url-regex/issues/34

Let's see what they say :)

davibe:
did they say anything ?

averissimo:
seems like there is already a pull request, but it's waiting to get merged :)

davibe:
if it does not get merged you can create your own fork of that lib and we can use that as a dependency :)

averissimo:
seems right!

davibe:
@averissimo what's the status on this ?

averissimo:
I was just revisiting this issue and it is not as straightforward as one would think.

There is definitely some cases that we can avoid, such as one or more parenthesis in the end of the url, but even then it might fail 

Test in hangouts web (not yakyak)

image

Test in github

(works http://sapo.pt/)
(works http://sapo.pt/something)
(works http://sapo.pt/aa(123)aa)

averissimo:
I think we can try and use google's method, remove all parenthesis in the end, unless there is an opening parenthesis in the path

averissimo:
I forked the project and managed to add the requirements to the regular expression.

Runs all previous tests and [added some](https://github.com/averissimo/url-regex/blob/master/test.js#L4-L20) based on past 2 comments in this thread :)

@davibe should I create an actual npm module or can it be installed from github in `package.json`

davibe:
For yakyak just referring to the fork is ok. For the sake of the community using url-regexp i don't know. Npm modules are owned by each publisher so you would publish a duplicate.. maybe you would need to change its name.. messy.

I'll just merge this

Issue #585, opened 2017-02-08
Online presence dot in dark theme is very hard to see


classicboss302:
It needs to stand out more, rather than being black on a dark background. It could be a colorful border around it or making it stand out more on its own.

averissimo:
@classicboss302 can you propose a pair of colors (inside and border) :)

classicboss302:
I would lean towards using the original colors as that keeps things consistent and would stand out fairly well. I think a border would only be needed if the color wasn't changed to something that stood out. In that case, a white border would provide contrast and help it stand out.

averissimo:
How about

Normal sidebar

![image](https://cloud.githubusercontent.com/assets/211358/22743885/dc1c7662-ee1b-11e6-9cee-c253ee30b138.png)

Hover/Selected conversation

![image](https://cloud.githubusercontent.com/assets/211358/22743901/ee5b137e-ee1b-11e6-9cfa-be6f220d612f.png)

classicboss302:
That looks good to me. Much easier to see.

averissimo:
pull request #586 

thanks for reporting :)

PR #586, opened 2017-02-08
adds online color as dark scheme does not have enough contrast

averissimo:
#585

- corrects indentation in global colors
- added `--online` color
  - All other color scheme inherit `--green` color, dark scheme inherits `--yakyakgreen`

### DIFF
--- a/src/ui/css/yakyak/colors.less
+++ b/src/ui/css/yakyak/colors.less
@@ -1,25 +1,27 @@
 // colors
 html{
---yakyakgreen: #0f9d58;
---white: #ffffff;
---ghostwhite: #f4f4f4;
---lightgrey: #e8e8e8;
---altgrey: #C5CED4;
---grey: #b4b4b4;
---darkgrey: #777777;
---lightblack: #455a64;
---black: #263238;
---green: #0f9d58;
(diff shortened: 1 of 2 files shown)


src/ui/css/yakyak/colors.less
```diff
@@ -1,33 +1,35 @@
...
+  --white: #ffffff;
+  --ghostwhite: #f4f4f4;
+  --lightgrey: #e8e8e8;
+  --altgrey: #C5CED4;
+  --grey: #b4b4b4;
+  --darkgrey: #777777;
+  --lightblack: #455a64;
+  --black: #263238;
+  --green: #0f9d58;
+  --online: var(--green);
```
averissimo: __new__

src/ui/css/yakyak/colors.less
```diff
@@ -36,6 +38,7 @@ html{
       --lightblack: #475866;
       --black: #A2A2A2;
       --green: #161c20;
+      --online: var(--yakyakgreen);
```
averissimo: exception...

note: `--green` itself could not be redefined as it sets the color of the top bar (that would be a major break)

PR #587, opened 2017-02-08
 Display timestamps in local locale (again)

nerijus:
I sent PR with moment.locale(window.navigator.language) a few months ago, and timestamps were displayed in my locale ("Šiandien 17:59" instead of "Today at 5:59 PM").
Does anyone know why it was changed? With moment.locale(i18n.getLocale()) timestamps here are again "Today at 5:59 PM", which I don't want.
OS is Linux.

### DIFF
--- a/src/ui/views/messages.coffee
+++ b/src/ui/views/messages.coffee
@@ -129,7 +129,7 @@ module.exports = view (models) ->
         div class:'historyinfo', ->
             if c.requestinghistory
                 pass 'Requesting history…', -> span class:'material-icons spin', 'donut_large'
-        moment.locale(i18n.getLocale())
+        moment.locale(window.navigator.language)
 
         last_seen = conv.findLastReadEventsByUser(c)
         last_seen_chat_ids_with_event = (last_seen, event) ->


averissimo:
it depends on the language of the interface. Yakyak now supports 8 languages.

`i18n.getLocale()` gets the current choice of language for the interface.

Maybe you can contribute with a new translation, there are only 117 strings that need translation ;)

nerijus:
But with my change it works even without translation, so why this change was needed again?

averissimo:
To answer your question, removing `i18n.getLocale()` won't have an effect in the rest of the interface, only the timestamps.

But that creates an inconsistency in the interface, with two languages being displayed. It would be strange to have the interface in german and then timestamps in portuguese.

I think the language options should determine the whole interface.

image

There's a workaround if you really wish to have the interface in english, but timestamps in other locale. Copy the empty `en.json` from the locale folder to the language that you wish _(example `lt.json` for lithuanian)_ and change the `__MyLocaleLanguage__` string from `English` to `Lithuanian`

This will keep the default strings (in English) and set the locale for the timestamps to the language code of the json file.

averissimo:
Another more permanent alternative would be to create an option to force the timestamps to be in the window navigator language. _(rephrase the name as you want)_

__I would merge that__, although I would prefer you develop a translation for YakYak to your prefered language

ps. This is the file that would need to be translated: https://github.com/yakyak/yakyak/blob/master/src/locales/en.json.template

nerijus:
Let's think about it:
1. Usually people will choose language which is the same as their locale. So my change does not affect them.
2. With my change it works for people which don't have translations in their language.
3. IMHO there are much less those, who will have "the interface in german and then timestamps in portuguese" than people which don't have translations at all.

So I vote for my change, or propose to add an item in Language menu called "don't change", "system language" or whatever.

nerijus:
Thinking even more about it, I rememeber on Windows (Linux too) there are translated apps, which display time/date in local locale, which may not be the same as the translation language. So I propose not to make yakyak code poluted, and don't even add "unspecified" Language item. If there are complains (which I doubt), just tell the user to change locale.

averissimo:
1. and 3. are kinda connected, if using the same locale as the operating system, then yes it does not affect them. But I usually switch the interface of apps depending on the country I am in, so now I'm using the Deutch translations. I tried your change and timestamps appear in Portuguese... seems strange :\

2. is a good point, but I'm not sure if an incomplete translation is better than just assuming an English/other fallback

I don't think the code as is should be merged, but change the pull request to add an option to `use system language for timestamps` or whatever and I'll merge it.

ps. we have conflicting usage expectation for YakYak, let's find a configurable middle ground :)

nerijus:
I actually don't care about translations, what I care is 24 hour clock. I don't like AM/PM style which I get without my change.
And again, I don't think a new option is needed.
For your case (which I doubt is very usual, and even then there is no problem imho in portuguese timestamps), you can just launch yakyak with "LC_ALL=de_DE yakyak" and it will display timestamps in German.

averissimo:
I'm gonna step back from further commenting on this pull request, I've voiced my opinion and whatever the community decides is fine by me.

davibe:
@averissimo maybe we could detect when the user locale is not supported by our translation and allow it to use its own locale for dates ?

davibe:
I have created a branch where the user can select as an option to translate dateformats or to use the system locale date format. It's simpler than handling a state where the user did not select any specific language as it was suggested before.

Currently the menu option is a checkbox and its called "Localize date format"

https://www.dropbox.com/s/fd3bubmy0q5zjga/Screenshot%202017-03-14%2017.13.00.png?dl=0

I would like for you to help me pick the best label for it
- Use system date format
- Translate date format
- Also translate dates
- better ideas ?

nerijus:
For me "Use system date format" is the best, "Localize date format" is second.

averissimo:
@davibe lol!! clearly all languages are necessary!! :p 

Agree with @nerijus choices

PR #604, opened 2017-02-21
Open obfuscated links directly

### LINKED ISSUE #593: URL parsing problem
I paste URL http://www.maps.lt/map/default.aspx#q=Pociuno%20g, then I click it in yakyak, http://www.maps.lt/map/default.aspx is opened.

averissimo:
fixes #593

### DIFF
--- a/src/ui/views/messages.coffee
+++ b/src/ui/views/messages.coffee
@@ -43,32 +43,40 @@ onclick = (e) ->
     e.preventDefault()
     address = e.currentTarget.getAttribute 'href'
 
+    #
+    # If a link starts with google.com/url?q=...
+    #  then it has been a masked URL by google.
+    #  We try to decode the original link and open it directly
+    # Otherwise, it could be a link only valid for the current logged user
+    # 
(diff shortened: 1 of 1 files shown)


src/ui/views/messages.coffee
```diff
@@ -43,32 +43,40 @@ onclick = (e) ->
...
-        return if e.target.status is 0
-        return if xhr.readyState isnt 4
-        finalUrl = xhr.responseURL
-        shell.openExternal(finalUrl)
-        xhr.abort()
-
-    xhr.open("get", finalUrl)
-    xhr.send()
+        finalUrl = fixlink(address)
+        shell.openExternal finalUrl
```
davibe: if you do `openExternal` here you also do it later after the xhr response. Something must be wrong

davibe:
this is not right. If the url starts with that pattern (www.google...) we have to first call the XHR to get the final address. You are bypassing that by doing `shell.open..` early

davibe:
i fixed it in another way see last commit https://github.com/yakyak/yakyak/commit/2b1defdaa2e9599f0b18c47443f72745d34ec0bf

Issue #597, opened 2017-02-18
Windows 8.1 /w 1.4.2 - volume mixer icon oversized


draeath:
As you can see in this screenshot, the icon is rendered oversize in the volume mixer, leading to some... interesting effects when YakYak is running.

![2017-02-18_1205](https://cloud.githubusercontent.com/assets/987449/23095118/a5b129ec-f5d2-11e6-9165-cee0cbfd19f0.png)

averissimo:
lol

averissimo:
This has been fixed in newest build of 1.4.2

I've tested in Windows 7, please let us know if it is solved

ps. using 32px icon in Windows (we were previously using 512px)

PR #607, opened 2017-02-24
uses 128px icon in tray

### LINKED ISSUE #600: Notification problem with systray
Hi,
in fedora 25 with classic Gnome, when I receive a new message I can see the pop-up notification and the systray icon becomes green. The problem is that in gnome classic, the systray did not show up even when the systray icon is green,
In the same situation, with pidgin, the systray icon has a blinking icon and the systray is not hiddden.
Last but not least: the icon in systray is really really small... too small.
Hope you can fix it, I really like yakyak.
M

averissimo:
fixes #600


averissimo:
added fix to #597 

going to merge as these are simple changes in code

Issue #606, opened 2017-02-23
New Language - Hebrew


ozbarshalom:
Hello,
This issue is to mention a new translate for YakYak - Hebrew

ozbarshalom:
I have already translated YakYak to Hebrew, which branch should I push this to ?

averissimo:
master!

ozbarshalom:
I don`t have any permissions to push to Master branch..

averissimo:
Ohhh sorry I misread your question.

You don't have permission to any branch, __you should create a pull request__ against master

See section "Creating the pull request" of  [this article](https://help.github.com/articles/creating-a-pull-request/#creating-the-pull-request) if you don't know how to.

PR #608, opened 2017-02-24
added hebrew language

ozbarshalom:
#606 - Add new language - Hebrew


averissimo:
thanks for the work, it is already available in latest build from v1.4.2 (pre-release)

averissimo:
oh I forgot to ask:

Can you add Hebrew to the README.md? (and probably take out the flag emojis as they are using unnecessary space)

PR #614, opened 2017-03-03
Add menu checkbox to control auto load on system startup

### LINKED ISSUE #478: Option to start yakyak automatically
I'd live to see an option to start yakyak up automatically when signing into the OS.

stdavis:
Closes #478

Go easy on me. I’ve not had much experience with coffee script or trifl. Let me know if I could do something in a better way.


src/ui/util.coffee
```diff
@@ -191,8 +192,13 @@ insertTextAtCursor = (el, text) ->
         range.text = text
         range.select()
 
+autoLauncher = new AutoLaunch({
+    name: 'YakYak',
+    path: '/Applications/YakYak.app'
```
davibe: the user may keep the app in a different location, right ?

stdavis: Ah yes! I forgot to remove that parameter. I've removed it in a subsequent commit.

I've also tested it on windows server 2012 and it seems to work fine now.

stdavis:
- [ ] Update [this wiki](https://github.com/yakyak/yakyak/wiki#add-to-application-menu) page when this branch is included in a release

averissimo:
Confirm that it works...

My only concern _(that does not affect the majority of users)_: I tried it using a development execution and it adds a barebones electron item to the startup _(see screenshot in bottom)_

if this could be detected and aborted with a warning message it would be awesome!!

```
notr 'Warning: <some message>' # this could be used to show a message
```

image

averissimo:
humm I'm really not sure if this is necessary, otherwise this feature cannot be tested and it does not automatically add on startup _(if option is on)_

what do you think @stdavis

stdavis:
@averissimo I also noticed the generic Electron app that's added when running in development. I think that we should leave it as is because that's the only way to test it in that environment. I think that developers can handle that. Perhaps I could add a note about this quirk in the wiki or something?

Also, I just noticed that the tests are failing on Travis. `AutoLauncher` is choking because the tests aren't run via electron. I'm currently testing code to work around this. I don't like adding branches in source code for tests but I'm unsure how to do it otherwise. I'll put a commit shortly and you can decide if it's the right way to go or not.

stdavis:
I acutally think that it's kind of nice to have the generic app in development. It helps you keep it straight from the prod release on your machine.

stdavis:
@averissimo This is ready from my point-of-view. Let me know if you want me to change anything else. Thanks for the fun project.

src/ui/util.coffee
```diff
@@ -191,8 +192,19 @@ insertTextAtCursor = (el, text) ->
         range.text = text
         range.select()
 
+# AutoLaunch requires a path unless you are running in electron/nw
+vesrions = process?.versions
+if versions? and (versions.nw? or versions['node-webkit']? or versions.electron?)
+    autoLaunchPath = undefined
+else
+    autoLaunchPath = process.cwd()
```
averissimo: why not use `process.execPath()` as the default in auto-launch?

image

stdavis: I think that you're correct. That's a more appropriate method. I've updated the code to use `process.execPath`.

averissimo:
Another todo (I'll do this one)

[ ] Add new string to template.json

PR #643, opened 2017-04-03
Louder 'sending' message indication.

lharding:
Subjectively I find it's a little hard to tell when outgoing messages are still in transit (especially important when the other conversant(s) are on something slow like SMS). Here's my proposal for making that indication a little louder.

### DIFF
--- a/src/ui/css/yakyak/messages.less
+++ b/src/ui/css/yakyak/messages.less
@@ -131,6 +131,12 @@
                 &.placeholder {
                     opacity: 0.5;
                 }
+                &.placeholder:before {
+                  position: absolute;
+                  bottom: 100%;
+                  color: pink;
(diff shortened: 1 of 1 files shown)


averissimo:
is it supposed to be like this?

image

maybe a little too faded :)

I like the idea, it is a lot more friendly

lharding:
That's much lighter than my build appears. Curious. On the bottom might be a good idea, especially if that gets this merged. Thoughts folks?

averissimo:
Something on the implementation side: it should use (or add) a theme-based color, see https://github.com/yakyak/yakyak/blob/master/src/ui/css/yakyak/colors.less

HomerSp:
Closing because there have been no updates in years.

PR #654, opened 2017-04-13
Use custom window frame for Linux/Windows platforms

Sazpaimon:
This is my attempt at fixing #211. Here's how it looks on Windows/Linux:

image

Clicking the hamburger will bring up the application menu as a context menu rather than as a separate row altogether.

I had to use SVG images from https://github.com/Templarian/MaterialDesign for the minimize/maximize/restore/close icons since Google doesn't include those in its font set. If you prefer a different way of integrating these icons, let me know.

This is my first time working in Electron so feedback is appreciated.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -127,6 +127,7 @@ app.on 'ready', ->
         icon: path.join __dirname, 'icons', icon_name
         show: false
         titleBarStyle: 'hidden-inset' if process.platform is 'darwin'
+        frame: false unless process.platform is 'darwin'
         # autoHideMenuBar : true unless process.platform is 'darwin'
     }
 
--- a/src/ui/css/yakyak/colors.less
+++ b/src/ui/css/yakyak/colors.less
@@ -50,9 +50,13 @@ html{
     .listhead button,
     .convhead .name,
     .controls .button,
-    .convhead .headwrap > .button,
+    .convhead .headwrap .optionwrapper > .button,
     .convadd .button{
       color: var(--black);
+    }
+
+    .convhead .win-buttons button{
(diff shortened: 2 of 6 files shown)


davibe:
I tested on mac and brifly on windows. I hope linux is fine. I think i will include this in 1.4.3. Thank you for your great contribution.

Just 2 comments for the next time
- i don't like code lines starting with `,`
- i prefer spaces before `{` in css

Sazpaimon:
Thanks! I also don't like leading commas and no space before `{` but I was following existing convention. I'll keep it in mind next time.

nerijus:
Using Window Maker on Linux it is OK:
screenshot-20170415_105338
But on MATE the window has no borders:
ekrano kopija 2017-04-15 10-55-41

nerijus:
I see it fixes issue 211 by hiding window borders. Is it possible to have an option which when selected would not hide the borders?

Sazpaimon:
If it makes more sense to keep the window frame on Linux, that can certainly be arranged.

Out of curiosity, how does [Discord](https://discordapp.com/apps) look on your setup? That was what I was trying to emulate here.

nerijus:
I don't have an account, without logging in it looks like this:
ekrano kopija 2017-04-15 17-44-11

nerijus:
Yakyak screenshot on another PC (MATE) - as you see, both left and bottom have no borders and is indistinguishable from the browser window below:
 
ekrano kopija 2017-04-15 17-50-46

Sazpaimon:
Okay, I think it makes sense to show the window border on Linux since there's no telling how a window manager will render a window. I'll make another PR to fix that.

PR #658, opened 2017-04-15
Only hide frame on Windows

Sazpaimon:
Based on some conversation on #654, I no longer hide the frame on Linux. It's just not possible to predict how a window manager will handle rendering a frameless window. This essentially makes Linux look like it did on 1.4.2, but I kept the menubar as a context menu as I thought it looks better than displaying new row altogether.

Tested on MATE (Fedora), Gnome (Fedora), LXDE (Ubuntu), XFCE (Ubuntu), KDE (Ubuntu), and Ubuntu Unity

Also removed a console.log call I forgot about.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -129,7 +129,7 @@ app.on 'ready', ->
         icon: path.join __dirname, 'icons', icon_name
         show: false
         titleBarStyle: 'hidden-inset' if process.platform is 'darwin'
-        frame: false unless process.platform is 'darwin'
+        frame: false if process.platform is 'win32'
         # autoHideMenuBar : true unless process.platform is 'darwin'
     }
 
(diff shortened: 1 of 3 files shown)


davibe:
thank you @Sazpaimon

chazz422:
Thanks for the fast update.  The links on the download page for 1.4.4 all say 1.4.3.  Are they just named incorrectly or are they the wrong version?

chazz422:
Just tested it and it works great.  Thanks.

averissimo:
@chazz422 

version was incorrect in `package.json` file, just bumped it in master to v1.4.4

Travis is working its magic to deploy new pre-release as I'm writing this

criscodex:
@averissimo What are the technical differences between pre-release and release? What's stopping v1.4.4 from becoming latest release? Thanks for your answers in advance.

averissimo:
nothing at all, for releases we just change the 'type' of release and that's it

pre-releases have new bugfixes and features and we hold on for a while just to make sure nothing breaks, but there's no fix protocol

PR #674, opened 2017-04-21
Add swedish language

Sjoo90:


src/locales/sv_SE.json
```diff
@@ -0,0 +1,173 @@
...
+                "show_window": "Visa fönster",
+                "toggle_minimize": "Växla minimerat till aktivitetsfält",
+                "start_minimize": "Starta minimerat till aktivitetsfält",
+                "close": "Stäng till aktivitetsfält"
+            },
+            "escape": {
+                "title": "Escape beteende",
+                "hide": "Göm fönster",
+                "clear": "Rensa fält",
+                "default": "default when tray is not showing"
```
averissimo: missing translation

Sjoo90:
Now added missing translation.

averissimo:
very nice work!! I always enjoy looking at a language that I don't understand :)

averissimo:
pre-release should be updated in 10-20 minutes with your work

averissimo:
arghhh, I forgot one thing, the language name should be in swedish instead of english, to be in line with other translations

Can you confirm that is 'Svenska'?

Sjoo90:
Yes that's correct

averissimo:
cool, i've pushed it directly and new binaries have been built! :)

it is already available in pre-release v1.4.4

PR #718, opened 2017-07-03
Add pt_BR support

lai0n:
Since portugal portuguese is not brazil portuguese


averissimo:
lamento ter demorado tanto tempo :)

bom trabalho!

Issue #232, opened 2016-06-17
Contact Names + Photos missing for SMS messages


derekderaps:
When I get SMS messages in my Hangouts, the name and photo of the contact don't show up (they do in the regular Hangouts app). It's not a broken image link, just the regular hangout icon,

![default hangouts icon](https://deraps.s3.amazonaws.com/f/photo.jpg)

Any thoughts? Thanks!

framerate:
Yeah, the name just shows as the phone number. Is this an API restriction?

davibe:
I guess messages in this case have a different payload.. never looked as i don't have friends sending sms to me on google hangout

panoptican:
Might be a Google Voice-specific feature now that that service can be integrated in Hangouts. To test, would suggest setting up a free Google Voice account and sending a few SMS to that number from your own phone. I'll probably give fixing it a try later this evening too.

adriangb:
I use Google Voice. Definitely a problem with SMS sent to the GV number.

davesimon:
I can verify this as an issue. I’ll look into the regular Hangouts web app to see what XHR requests it’s making. I’m using a Google Voice number, which for a long time the Hangouts app had the same trouble with. I ended up re-inputting many people into Contacts to get it to work.

jacobmischka:
Same here, though I'm using Project Fi with a ported Google Voice number. Though I'm pretty sure the backend behavior of Fi and regular GV are the exact same, just commenting for completeness's sake.

mattrock5a:
BUMP... would love to see this fix implemented, and know many others who say the same!

underscorephil:
+1

bwiernik:
Also, messages sent to/from the GV number don't display the "SMS" tag over the contact picture.

dinoboy197:
Confirmed to still be an issue in 1.3.3. This is a weakness over Google's official client.

AndrewJDR:
Is this a restriction of hangupsjs? Should an issue be logged there as well?

averissimo:
**We need someone that uses hangouts with sms to debug and find why**

it could be a problem with the library, or maybe the name and photo have a different structure with this type of contact

if someone wants to pick this up I can help showing in which data structures to look

AndrewJDR:
@averissimo Sure, if you want to tell me what to look for (what exact commands to run, what data to look for, etc) I'll try to help.

averissimo:
[...]

Issue #90, opened 2015-12-01
Sent SMS texts stay gray


kethinov:
I'm using YakYak as a client for my SMS conversations with [Project Fi](http://fi.google.com) integrating my phone number into hangouts.

This enables text messages sent to my phone number to be routed through my hangouts account. I receive texts in YakYak.

I can reply (again routed through Project Fi) but the replies stay as gray text instead of showing as fully black text, indicating some kind of network error.

However, there is no network error and the texts do get sent correctly when I verify on my phone.

swrobel:
:+1:

algesten:
the gray is until we get some kind of confirmation back from the server. perhaps we never get that confirmation for these messages?

jeffwllms:
it doesn't as if the recipient receives the message.  I have he same issue.  sending to someone on project fi.  They are grey on my end and show on my phone.  But the receiver doesn't see them (on project fi).  I receive messages they send.

dinoboy197:
Confirmed as still an issue in version 1.3.3. I would love to use this client, but unfortunately, this problem prevents me from being able to switch from Google's client.

natecox:
I'm having this issue now as well. Is this on the radar for being fixed?

davibe:
@natecox There is no radar, just issues and pull requests. I don't see any PR about this so for now it's a no

chazz422:
This is a major problem with this app right now.  I've had several clients complain about not receiving my text messages.  Only happens with yakyak.  From hangouts client, it works fine.

Yes, it's strange how it shows up on the hangouts client on my phone as if they were sent, but they never get to the destination.

natecox:
This is certainly still an issue, commenting to prevent automated closing.

averissimo:
@natecox, can you help test a pull request that might address this issue?

natecox:
@averissimo Sure, which one?

chazz422:
I can help test also.

averissimo:
#738 or the `hawksprite-feature/google-voice-sms` newly created in yakyak/yakyak

For a binary file in 5 to 15 minutes there should be a release called v1.4.4-google-voice

__For this to work you need to__:

- activate an option called `Google Voice Mode` in main menu
- try to send an sms in a google voice conversation

note: this changes the way all messages are being sent in the client to use "google voice"... I'll try to change this if the pull request is working, but I'll need someone to test this out as I don't have access to a google voice account.

chazz422:
[...]

PR #738, opened 2017-08-20
Feature/google voice sms

hawksprite:
This adds the capability for users to use yakyak while also wanting to send google voice SMS through a related hangouts conversation.

Internally the current YakYak setup is to always use the BABEL delivery method.  When you attempt to post a message to hangouts in a conversation that is using GOOGLE_VOICE it fails.  This adds a toggle option in the context menu as a way for user's to change the default behaviour of the library's delivery method.

NOTE:  This is probably seen as a "quick fix" for the issue, a full fix would be automatically detecting the delivery method for each conversation at a lower level.

Also, for Google Voice users setting the delivery method to GOOGLE_VOICE doesn't make BABEL conversations fail to send for some reason.  So this is more of a check it if you're mostly a google voice user.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -252,12 +252,19 @@ app.on 'ready', ->
     # sendchatmessage, executed sequentially and
     # retried if not sent successfully
     messageQueue = Q()
-    ipc.on 'sendchatmessage', (ev, msg) ->
+    ipc.on 'sendchatmessage', (ev, msg, googleVoice) ->
         {conv_id, segs, client_generated_id, image_id, otr, message_action_type} = msg
         sendForSure = -> Q.promise (resolve, reject, notify) ->
             attempt = ->
                 # console.log 'sendchatmessage', client_generated_id
                 delivery_medium = null
+
+                ## If the client isn't in google voice mode null is fine, otherwise specify google voice
+                if googleVoice
+                    delivery_medium = [2]
+                else
+                    delivery_medium = null ## Retain the null status to let the library upstream decide, currently this will always be BABEL
+
                 client.sendchatmessage(conv_id, segs, image_id, otr, client_generated_id, delivery_medium, message_action_type).then (r) ->
                       # console.log 'sendchatmessage:result', r?.created_event?.self_event_state?.client_generated_id
                     ipcsend 'sendchatmessage:result', r
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -316,15 +316,22 @@ app.on 'ready', ->
     # , false
 
     # we want to upload. in the order specified, with retry
-    ipc.on 'uploadimage', seqreq (ev, spec) ->
+    ipc.on 'uploadimage', seqreq (ev, spec, googleVoice) ->
         {path, conv_id, client_generated_id} = spec
         ipcsend 'uploadingimage', {conv_id, client_generated_id, path}
         client.uploadimage(path).then (image_id) ->
-            client.sendchatmessage conv_id, null, image_id, null, client_generated_id
+
+            delivery_medium = null
+            if googleVoice
+                delivery_medium = [2]
+            else
+                delivery_medium = null
+
+            client.sendchatmessage conv_id, null, image_id, null, client_generated_id, delivery_medium
     , true
 
     # we want to upload. in the order specified, with retry
-    ipc.on 'uploadclipboardimage', seqreq (ev, spec) ->
+    ipc.on 'uploadclipboardimage', seqreq (ev, spec, googleVoice) ->
         {pngData, conv_id, client_generated_id} = spec
         file = tmp.fileSync postfix: ".png"
         ipcsend 'uploadingimage', {conv_id, client_generated_id, path:file.name}
@@ -333,7 +340,12 @@ app.on 'ready', ->
         .then ->
             client.uploadimage(file.name)
         .then (image_id) ->
-            client.sendchatmessage conv_id, null, image_id, null, client_generated_id
+            delivery_medium = null
+            if googleVoice
+                delivery_medium = [2]
+            else
+                delivery_medium = null
+            client.sendchatmessage conv_id, null, image_id, null, client_generated_id, delivery_medium
         .then ->
             file.removeCallback()
     , true
--- a/src/ui/dispatcher.coffee
+++ b/src/ui/dispatcher.coffee
@@ -124,10 +124,10 @@ handle 'selectConvIndex', (index = 0) ->
     viewstate.selectConvIndex index
     ipc.send 'setfocus', viewstate.selectedConv
 
-handle 'sendmessage', (txt = '') ->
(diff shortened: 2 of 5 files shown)


averissimo:
Following  #90 discussion:

- What was appearing before the question mark?
- Was sending an image/mms working before?

averissimo:
We need some information from an google voice conversation.

We just released a new v1.4.4-google-voice binary

Can someone open Yakyak and:

- Open a google voice conversation
- Press Ctrl + Alt + i to open the developer pane
- Open the Console tab 
- Expand the last conversation / outer / inner and screenshot it  as the picture below shows

This should not reveal any unique id or something to compromise privacy, if it appears please censor it :)

screenshot from 2017-09-14 18-40-55

natecox:
image

averissimo:
great, so it's possible to get the delivery medium from the conversation itself :)

Working with this will avoid having a global option,

@hawksprite hope you don't mind if I change a couple of things in your code :)

An implementation in the library would be better, but due to the asynchronous nature of the library, it might be best to manage this in yakyak

hawksprite:
I agree, feel free!  This pull was meant more as a way to outline the core issue with the google voice delivery.

averissimo:
@natecox 

- What was appearing before the question mark?
- Was sending an image/mms working before?

__Can you compile yakyak from source?__

I've adapted the code in the branch hawksprite-feature/google-voice-sms to detect the delivery medium and use it, instead of using a global option. 

I'd like if someone test it? _(currently only for text sms.. for images it fallbacks to as it is in normal releases)_

chazz422:
I've never had the google voice SMS icons working.  Always a question mark for me.  I only get icons for hangout messages.  Also, I don't get names, only phone numbers for SMS.

As for image/mms, I haven't tried it through SMS.  I only send images to hangouts users.

natecox:
@averissimo The icons have never worked for me on YakYak, I've always seen them as `?`. Same with image sending, although that's not something I do often anyways.

averissimo:
Ok, cool I just don't want to break functionality with this new code, so I was making sure :)

The binaries are being built for you guys to test if __sending text__ to google contacts works out of the box _(without any global option)_

Can you guys test: https://github.com/averissimo/yakyak/releases/tag/v1.4.4-google-voice-av

chazz422:
Global menu option is still there.  Is that how it is expected to be?

averissimo:
Forgot to delete it :) but that was no longer used for the actual sending of messages

anyway, a new version at that link should be available without that option and some leftover ui code

Can you give feedback on it?

chazz422:
Yes, it's working without the global option checked.

averissimo:
Ok, keep testing it in your day to day :)

I'll prepare the merge as this requires the base library to also be updated

chazz422:
Will do.  By the way, I'm running on Linux, but don't think that should
matter for this type of change.  I have a windows machine at home that I
can try it on later.

averissimo:
@hawksprite do you want to merge the yakyak/hawksprite-feature/google-voice-sms branch?

or I can create a new pull request and change this?

This is basically the same so if you don't say anything by the time hangupsjs library is updated I'll do the latter

averissimo:
@chazz422 it should not, but the more widely tested this is the better :) so we don't have any surprises ;)

nsubtil:
The binaries from https://github.com/averissimo/yakyak/releases/tag/v1.4.4-google-voice-av seem to work for me --- GV messages now get sent correctly.

averissimo:
@nsubtil great! thanks for testing

This will be merged and released soon

lazyb5:
Are you guys going to try and get contact names instead of just showing the phone number? Like the chrome hangout extension.

averissimo:
Going to merge branch where newer work was done.

For more information see newer pull request that I'll create in a moment

averissimo:
Can someone test pre-release v1.4.4 ? it should implement all the changes discussed in this pull request correctly.

We are driving for a new main release after testing this one out :)

Cheers and thanks again for the feedback and contributions!

plessbd:
This new version allows me to send messages to my sms contacts via project fi!

So from my testing this works wonderfully

lazyb5:
I'm using hangouts with google voice and contact names still aren't sync. It just shows the phone number. The chrome extension shows contact names.

averissimo:
@lazyb5  that is not covered in this new feature in yakyak :\

The data we get from hangouts is just the phone number that is shown.

Something else must be done to obtain the contact names, you can check out some discussion on this on #232

lazyb5:
@averissimo OK thanks, I'll keep checking to see if any progress is made. Thanks to all the devs this is a great program. I would use over the extension if I wasn't a voice sms user.

chazz422:
I've installed v1.5.0 and have a couple problems that were not in 1.4.3.

On the windows platform, the delete conversation option for SMS messages is not working.  They just stay there as if nothing happened.

On the Linux platform, deletion works fine, but when a new SMS arrives that was previously deleted, it will crash yakyak.  If I leave the SMS without deleting it, then new messages will not cause the crash.

Not sure why it behaves so differently on each platform but hopefully you can reproduce this.

averissimo:
I don't use sms so it's difficult to reproduce, is this something that only happens to sms conversations or also occurs with normal hangouts?

Can you run yakyak from the command line and check if some exception appears?

chazz422:
@averissimo I usually don't delete hangouts, only sms, but I tried, just now, to delete an old hangout and it wont delete in windows.  So it appears the problem is not isolated to just SMS.

averissimo:
what do you mean deleting and hangout/sms.. is it going to "details" -> "delete conversation"

or something else?

Trying to reproduce on linux and it works

chazz422:
yes, deleting from the "details" -> "delete conversation".  Regardless of if it originated from google voice sms or native hangouts messaging, the conversation will not disappear on windows.

The "delete conversation" is working fine on linux, but if I receive a new voice SMS, it will crash the client.  For example, amazon will send me a SMS from +1262966 and it will crash when I click to view it.  After restarting the client, I can click and view the message without a problem.  As long as I don't delete the message, I can receive new ones.  But if I delete it and then receive another one, it will crash again.  I'll get you the details of the crash from the command line like you requested but I have to wait til I get another one.  I don't do much with SMS besides amazon notifications and my bank (which also has the same problem).  Just to reiterate, this crashing is only on linux version, not windows.

So 2 separate problems here, but don't know if there's any relation.  Neither of these problems existed in 1.4.3.

averissimo:
Can you run yakyak from the command line and report which exception appears?

chazz422:
I ran it from the command line and left the terminal open to monitor for exceptions.  I received a SMS for amazon delivery today and it didn't crash.  So the problem's not happening.  I'm going to try running it with nohup to capture tty output to a log.  I'll do some more troubleshooting as I can and let you know what I find.

averissimo:
Classic!! if so please open a new issue :)

Issue #245, opened 2016-06-20
Signing the binary?


YesThatAllen:
I, as would others, I'm sure, would be open to signing this project's binary. 

Let me know know if there are plans for that, or if there's a fund I can contribute to for the Yak to get its own dev account.

davibe:
I never looked into this but as yakyak becomes more and more popular i can see how it would help.

johnduhart:
Still relevant, for example my corporate policy prevents users from installing unsigned binaries.

danchr:
I've posted pull request #768 which should enable signing binaries when built on a Mac. What's the build process for releases? Are they built manually or by Travis?

averissimo:
They are built by Travis

averissimo:
Reopening just to see if any conclusion was reached. If not just let it close again.

PR #742, opened 2017-08-24
add support for signing the Mac application bundle (fixes bug #245)

danchr:
By default, use the 'Mac Developer' identity, which most people should
have. Please note that the build still succeeds if signing fails.

### DIFF
--- a/gulpfile.coffee
+++ b/gulpfile.coffee
@@ -72,6 +72,7 @@ deploy_options = {
         ProductVersion: "#{json.version}"
     }
     osxSign: true
+    osxIdentity: process.env.CODESIGN_IDENTITY or 'Mac Developer'
     arch:     archOpts.join ','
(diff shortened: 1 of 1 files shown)


averissimo:
Thanks for the changes, looks good but it still requires an identity or can we use the "Mac Developer"?

Looking at Travis-CI documentation, it's possible to use a OS X build environment. Do you think we can set up some kind of official identity/certificate to perform this?

_Background:_ we use Travis-CI.org to run all tests and build the binaries

ps. Did you try using the `electron-packager` default method for signing?

anywayWe are building the binaries using travis-ci

danchr:
> ps. Did you try using the electron-packager default method for signing?

Nope, I couldn't get it to work — I just discovered that it was caused by a missing dependancy. I'll open a new PR which fixes that.

Issue #529, opened 2017-01-04
Feature Request - provide contact info on new conversation page


thenextdon13:
Hello-
One thing that has been annoying with hangouts was the disassociation of contact info (specifically email address) and user names.

This was frustrating because duplicate names (i.e. the same person with accounts on 2 different domains) were nearly impossible to differentiate.  

I see more recent versions of hangouts  this (by providing the email address under the name), but yakyak does not do anything simlar.

So, if possible, please add contact info (email address) to contacts resolved in the 'new conversation' user search box.  

many thanks

thenextdon13:
Looks to be working, thank you!

averissimo:
You can thank @lonvarscsak for his great work!

PR #747, opened 2017-08-29
Fixed a bug in drawAvatar where email address was pulling in the wron…

lonvarscsak:
…g property (resulting in it being null).  This means that if there is a known email address it will appear as the "title"  Also unified avatar drawing in the convadd code  since there were differences between how things were drawn between selected contacts and unselected contacts (which was odd behavior).  This should fix Issue #529

### DIFF
--- a/src/ui/views/convadd.coffee
+++ b/src/ui/views/convadd.coffee
@@ -82,7 +82,7 @@ module.exports = view (models) ->
         ul ->
             convsettings.selectedEntities.forEach (r) ->
                 cid = r?.id?.chat_id
-                email = (r.properties?.email?[0] ? entity[cid]?.email?[0]);
+                email = (r.properties?.email?[0] ? entity[cid]?.emails?[0]);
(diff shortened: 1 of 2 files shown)


src/ui/views/convadd.coffee
```diff
@@ -82,22 +82,27 @@ module.exports = view (models) ->
         ul ->
             convsettings.selectedEntities.forEach (r) ->
                 cid = r?.id?.chat_id
-                li class: 'selected', ->
-                    drawAvatar(cid, viewstate, entity)
+                email = (r.properties?.email?[0] ? entity[cid]?.email?[0]);
```
averissimo: shouldn't here be also emails in the second one?:

`email = (r.properties?.email?[0] ? entity[cid]?.emails?[0]); `

lonvarscsak:
Hey André, thanks for the comment.  Yes, you are correct...good catch.  Fixed.

PR #761, opened 2017-09-13
adds (y) as shortcode for thumbs-up

### LINKED ISSUE #684: convert (y) to thumbsup emoticon before sending message
Hi, I just switched to yakyak and it's great.

Are you interested in converting some chars to their emoticon alternative before sending the message. For example in WhatsApp and the Google Hangouts webclient, (y) gets converted to :+1:.

Would be great if yakyak would mimic this behavior. I'm sure there are more to add but (y) is the one I use frequently.

Tnx for yakyak :smile:

averissimo:
closes #684


averissimo:
An ideal work on top of this would be to parse all non elements that are not official shortcodes (i.e. surrounded by `:`) and add that to the regular expression in util

robkorv:
I just fired up master and typing `(y)` doesn't get me a thumbs up :disappointed:

averissimo:
it is working in my machine with latest commit: 52f0757 

Are other smileys working? :) :\ :thumbsup: ?

robkorv:
Other smiley do work :+1:. `(y)` not.

averissimo:
This is super strange, I've just tried the latest pre-release and it's working.. _(i.e. downloading the binary from github)_ 

Using `(Y)` uppercase y does not work though.. could this be it 

sidequestion: do you think it should also be converted?

__To debug:__

- Have you checked the developer's pane (ctrl + shift + i) to see if there are any exceptions?

- Make sure that there is `(y)` shortcode in `src/ui/emojishortcode.coffee` (should be the last line)
  - and then make sure that src/ui/util.coffee also contains a string "[(]y[)]" in line 161

robkorv:
Ok, for some reason when I download the github version I get it. But when I start from master it doesn't work, while I don't have any errors and the `(y)` is in the source file.

I start it with `nohup npm run electron app &`

Uppercase `(Y)` also get's converted in google's webclient. I think it's a good idea to implement it.
Tnx for working on this, I think me not getting it from master is a problem at my end.

averissimo:
You are forgetting to compile.

I recomend you use `nohup npm start &`

This compiles and runs the app

robkorv:
Tnx! Now it works :+1:

PR #768, opened 2017-09-15
Mac signing tweaks

danchr:
This pull request supersedes #742 and contains a few tweaks:

1.  Add a missing dependancy so that signing works on a Mac. 
2. Specify the identity so that it matches a common, free-of-charge identity.
3. Set an [app category](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html#//apple_ref/doc/uid/TP40009250-SW8) — which the Finder can use when sorting applications by category.

I'm not sure which identity is available on your Travis setup, but this at least fixes signing on my Mac :)


averissimo:
check out the mac-tweaks branch in yakyak.. I'm trying to tweak `.travis.yml` to build the osx version in mac with your changes

travis build: https://travis-ci.org/yakyak/yakyak/builds

davibe:
I noticed the latest build is green. Then I downloaded `yakyak-1.4.4-mac-test` and I still get

screenshot 2017-09-16 09 58 26

averissimo:
:(

__was it really signed?__  I'm guessing signing with "Mac Developer" will only be recognized in the same computer it's being built, otherwise the purpose of signing is kinda defeated.

the logs are not very clear if it happened or not

I think that we might need a real certificate and something like this: https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-/tree/master/sig

danchr:
> **was it really signed?** I'm guessing signing with "Mac Developer" will only be recognized in the same computer it's being built, otherwise the purpose of signing is kinda defeated.

Unfortunately, it wasn't:

    $ codesign -vv ~/Downloads/YakYak-darwin-x64/YakYak.app
    /Users/dan/Downloads/YakYak-darwin-x64/YakYak.app: code object is not signed at all
    In architecture: x86_64

I'm guessing that there's no identity matching “Mac Developer” available. Just so we can check that a subscription isn't necessary, I've uploaded a [binary](https://github.com/danchr/yakyak/releases/tag/v1.4.4-mac-test) to my fork. That one is signed:

    codesign -vv dist/YakYak-darwin-x64/YakYak.app
    dist/YakYak-darwin-x64/YakYak.app: valid on disk
    dist/YakYak-darwin-x64/YakYak.app: satisfies its Designated Requirement

> I think that we might need a real certificate and something like this […]

Well, you *will* need a real certificate, but preferably one from Apple. It's not that hard, assuming you have a Mac: sign up at develop.apple.com, and then get the certificate using Xcode. I think code-signing certificates from other CAs work, but they tend to be quite expensive compared even to a paid Apple Developer account.

I think you should be able get by with a much simpler solution than what they have. I've added a commit that allows overriding the identity & keychain. Then, you just need to have Travis unlock a keychain and set two environment variables, and then it should Just Work™ 😉

danchr:
You could try setting the ``DEBUG`` environment variable in the Travis configuration:

    DEBUG=electron-osx-sign npm run gulp deploy:darwin && codesign -vv dist/YakYak-darwin-x64/YakYak.app

danchr:
Quoting myself:
> I'm guessing that there's no identity matching “Mac Developer” available. Just so we can check that a subscription isn't necessary, I've uploaded a binary to my fork. That one is signed:

I just tried downloading it, and the signature doesn't pass muster. In other words, you *do* need a paid subscription in order to be an “identified” developer…

averissimo:
If a paid subscription is necessary it really sucks! :\

Mac is not my arena, so I pass the ball to @davibe :p

danchr:
Coming to think of it, you might not want to have Travis sign all builds anyway… At the very least, it shouldn't sign pull requests automatically; that's a glaring vulnerability.

averissimo:
We are only building on tag updates, which are either releases or pre-releases

Travis only runs tests on other updates (such as pull requests, commits, etc..)

davibe:
Yes, it looks like we would need a paid apple subscription and i cant use the same iOS (mobile) I already have for desktop apps. At the moment I am not willing to pay extra for it.

HomerSp:
Since this seems to require a paid account, and there have been no updates on this in years now I'm going to close this pr.

PR #769, opened 2017-09-16
Respect ON_THE_RECORD / OFF_THE RECORD

### LINKED ISSUE #675: 'Do not save history' feature is not working with YakYak
In Hangouts chat it is possible to turn off saving of chat history. It seems that when the feature is on, hangouts just periodically cleans up messages from chat, so they can resist for just few hours or 1-2 days. However when these messages sent via YakYak they seem to stay forever. Maybe there should be some special flag or something. The chat setting is only available in Hangouts web/mobile app. 

See also https://github.com/yakyak/yakyak/issues/664

averissimo:
respects the otr status and shows message when it changes (i.e. recognizes the event)

closes #675


skawaii:
I just tested this out and it seems to properly respect the history being on/off.

One small bug I noticed was that when the history was turned back on, the message I saw was "History has been turned off". It should have said "on".

image

Having a more permanent notification (icon, changing the background color of the messages like Google does, or both) would be a very welcomed feature in the future.

averissimo:
corrected the off/on :) it was a typo

Do you want to take a crack at styling the "off the record"? :)

src/ui/dispatcher.coffee
```diff
@@ -421,6 +421,9 @@ handle 'handlesyncedevents', (r) ->
             conv.addChatMessage e
     connection.setEventState connection.IN_SYNC
 
+handle 'syncrecentconversationsnow', ->
```
pago88: what's the reason for creating this handler?

averissimo: For what I remember when the conversation goes on/off the record we get a conversation event from hangouts that signals the change, but not a actual message.

This only happens after re-syncing the last messages. I didn't want to create an exception when handling conversation_events so a trigger to the `syncrecentconversations` is performed, but it  has a 10second timeout before running

so `syncrecentconversationsnow` was born 🔥

I didn't try setting the previous handler with a default argument `..., (timeout = 10000) -> ...` so that it preserves compatibility

which could be a better solution, what do you think?

pago88: That makes sense. I just didn't see any calls to syncrecentconversationsnow action, so I'm wondering if there's any use for it if we're not calling it? Or maybe the part of the code that calls this was not committed.

averissimo: oh!! great eyes!!

Corrected it now.

pago88: I believe the call to syncrecentconversationsnow is causing an issue that shows up at least half of the time when sending a message. When you send a message, it will normally display the text  as light gray (while it's sending), and then actually display it as normal black text once it's sent. The issue that shows up here is that a lot of the times the text will show light gray, then disappear completely, and then show up as black text. The amount of time it disappears varies, from very short amount of time to very noticeable.

averissimo: Yup, I'm experiencing the same. I've started to work on something a little more intelligent, i.e. update the specific conversation instead of resyncing everything.

Also, the call should only be performed on specific situations and not always _(as it it now)_

Can you change the review to `request changes` ? so that it is shown as not ready?

src/ui/models/conv.coffee
```diff
@@ -24,7 +24,7 @@ add = (conv) ->
     # participant_data contains entity information
     # we want in the entity lookup
     entity.add p for p in conv?.participant_data ? []
-    later -> action 'syncrecentconversations'
+    later -> action 'syncrecentconversationsnow'
```
pago88: The call to syncrecentconversationsnow is causes an issue when sending a message where the text disappears for a short time between it's sending state and when it shows as black  text that has been sent.

averissimo:
Rebased code and removed `syncrecentconversationsnow`

There might be a delay with the status of the conversation changing in YakYak. (i.e. 10seconds before any messages sent from yakyak are OffTheRecord)

Can we live with this lag? @pago88 @skawaii @inliquid

skawaii:
I think I could live with that delay. It would obviously be better without the delay, but I'd rather have the delay than not have the feature at all.

By the way, my apologies for never responding to you asking if I wanted to help with styling "off the record" mode. I'm interested in helping, but can't commit any time right now. I'll let you know if that changes.

Thanks for all the work you guys put into this!

pago88:
@averissimo I think the delay should be fine. I was working on some code that integrated your changes from this pull request with an actual button for toggling OTR, as well as having some visual feedback of what the conversation's status is. But I got busy there these past few months. I'll go back this week and see where I left it off.

averissimo:
Did you introduce any form of notice that conversation was `OFF_THE_RECORD`?

pago88:
@averissimo I pretty much added a button to the conversation head that shows if history is on or off for that conversation, and clicking the button toggles history. I just rebased and tried out the code I had; it seems like something may have changed, but I'm not sure if it's something on my side of the code. If the history is changed in gmail, yak yak used to display a message in that conversation that said History has been turned on, correct? I will look into this and probably commit something tomorrow.

 What I was still working on was getting the conversation bubble's color to change, to mimic hangouts' behavior. I hadn't yet figured out how to make it work correctly tho.

averissimo:
yes, hangouts does show that message when turning history off. I think that when turning back on it does not send a message, but it activates a flag:

https://github.com/yakyak/yakyak/pull/769/files#diff-fd3df69ddfd73954ad96570fe5591e31R276

image

pago88:
@averissimo 
My initial commit is here https://github.com/pago88/yakyak/tree/toggle-otr-function
Lemme know what you think. Maybe we can just consolidate both branches into one?

pago88:
@averissimo 
I submitted a pull request #866 that contains my changes for the ability to toggle history on or off. It uses some of the logic from this PR

Igor2399:
waiting for #866 ))

PR #770, opened 2017-09-18
adds names to top of message groups

### LINKED ISSUE #753: Show names in group chats
I work for an organisation that requires all users to have the same display picture. Group chats are hard to use in Yakyak as it only shows display pictures, and no names. is it possible to add this functionality in further builds?

averissimo:
closes #753 

@davibe was not showing the names of the users a design decision?

I can make this code work only in group chats


averissimo:
selection_00938

Issue #223, opened 2016-06-16
Global Shortcut for Show/Hide


derekderaps:
v1.3.1
OS X

I love the option to run faceless (no Dock icon) but need a global show/hide shortcut to pull it back up without going to my system tray.

davibe:
I think it can be difficult to pick a generic shortcut for yakyak therefore it should be configurable (?)

acca:
v1.4.3
OS X

I am also interested in this feature. Usually I use `CMD + m` or `Cmd + h` to close or hide the chat window. Then I use `CMD + tab` to switch back to the application. Once the application has the focus I would like to open the chat window with a shortcut. I propose few options:

1) As of now `CMD + 1` brings to the first chat if you have the chat window already open. This should work also when the chat window is on hidden or in the background or even closed. This works in many other OSX applications: when you are in the application context and you press any particular shortcut the main window comes back to the foreground (Try with Mail.app -> `CMD + w`  to close main window and `CMD + 1` to with the Sent mailbox)

2) the function of `Toggle minimise to tray` in the system tray icon already does the trick. It is just inconvenient to use the mouse. Why not replicate the same menu item in the `View` or `Window` main menu of YakYak? In this way every user could map the preferred shortcut using `Keyboard->Shortcuts` in OSX Preferences

I will be happy to help testing any possible solution.

Issue #811, opened 2017-11-12
Menu entry Tray Icon confusing


JPRuehmann:
Hello
To maximize the Window I hav to click on minimize window to tray in the Tray Menu this is a little bit confusing.
Thanks

dustin-cowles:
I came here to report this same issue. The behavior of the tray/menu icon on mac should be the same as the system tray icon on Windows.

On Mac, when left clicking the icon it opens the menu and I must select "Toggle minimize to tray" (as an aside, it should say "Dock" as that is what it is really referring to)
On Windows, when I left click the system tray icon it toggles, when I right click I get the same menu, sans the toggle item.

averissimo:
@JPRuehmann can you suggest a new wording for this?

@davibe can you check this out on mac?

JPRuehmann:
restore Window (Fenster wiederherstellen)

PR #842, opened 2018-01-04
Add toggle main window menu item

acca:
- Solve issue #223

### DIFF
--- a/src/locales/it.json
+++ b/src/locales/it.json
@@ -171,6 +171,7 @@
             "title": "Finestra",
             "minimize": "Riduci",
             "close": "Chiudi",
+            "toggle_window": "Mostra/Nascondi finestra principale",
             "front": "Porta tutto avanti"
         }
     }
--- a/src/locales/it.json
+++ b/src/locales/it.json
@@ -99,7 +99,7 @@
             "hide_dock": "Nascondi l'icona della barra delle applicazioni",
             "fullscreen": "Schermo Intero",
             "tray": {
-                "show_tray": "Mostra icona nel vassoio di sistema",
+                "show_tray": "Mostra/Nascondi finestra",
                 "show_window": "Mostra finestra",
                 "toggle_minimize": "Attiva/Disattiva riduzione nel vassoio di sistema",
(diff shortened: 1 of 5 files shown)


src/locales/it.json
```diff
@@ -171,6 +171,7 @@
             "title": "Finestra",
             "minimize": "Riduci",
             "close": "Chiudi",
+            "toggle_window": "Mostra/Nascondi finestra principale",
```
averissimo: Can you also add this to `en.json.template` ?

src/ui/views/menu.coffee
```diff
@@ -438,6 +438,10 @@ templateWindow = (viewstate) -> [
         accelerator: getAccelerator('close')
         role: 'close'
     }
+    {
+        label: i18n.__ 'menu.window.toggle_window:Toggle window show'
```
averissimo: Could this also be done via the `Hide YakYak` in `File`? and changing its action to `toggle` *(instead of `hide`)*

 or is there a convention that window menu has this option?

Would `Toggle window show/hide` be a possible string? (I would prefer to change the "Toggle minimize to tray" to something like you suggest here, which could be used by both commands) -- related to #811

acca: Probably it can ben done also via the `Hide YakYak` menu but in OSX almost all applications have this (press `CMD + h` for hide an app) convention and I would suggest to not change that but instead add a specific menu item for that feature.

There is no such convention for submenu item in the `Window` menu, I just used that menu because seems appropriate. Of course I need a submenu item (anywhere) to control that feature with a keyboard shortcut in OSX Preferences (see https://goo.gl/wb59Du)

Yes that can be the right string for both the menu item and the system tray. The Italian could be simply changed in `Mostra/Nascondi finestra`

Feel free to change the PR or let me know your preferences I can do that.

averissimo: Thanks for the reply

I've sent a PR request to you that will trickle down to this one. I still haven't figured out if I can contribute to it directly, but this way it's better as it requires your approval :)

PR #845, opened 2018-01-05
Fix position when closing window in multi-monitor

### LINKED ISSUE #757: Multiscreen Issues
In Windows 10, I have two monitor setups. My home office has a monitor above the 'main' one. When I am remote I have a portable that I use. Yak Yak seems to save the absolute position on the virtual desktop and restore that.... even if that is well off of the existing physical screen space. Short term, could there be a setting to just reset the position on startup? (i.e. not save screen position from previous run. ) Longer term it would be nice if the window would just note that it is off of the visible screen on startup and move itself to somewhere where it can be seen.

### LINKED ISSUE #792: Memorizing last window position and size
I'm running yakyak on Ubuntu 17.10 with vanilla gnome on Xorg.
Since I have a multi monitor setup in 99% of the time the yakyak window appears on the splitting line of two monitors leaving it part on one and part on the other.

An option maybe to make it remember its last position and size would be very good.

averissimo:
closes  #757
closes #792


averissimo:
Got some positive feedback on this pull request, merging

PR #856, opened 2018-01-14
Make flatpak optional

isavegas:
As the `electron-installer-flatpak` npm package is only required for the MacOS and Linux flatpak targets, I moved the dependency to the `optionalDependencies` object. Unfortunately, there isn't an `optionalDevDependencies` construct.


averissimo:
Great work!

Unfortunately there are some dependencies needed such as `zip`, but that goes for every OS.

I would like to use `gulp-zip` but there is a bug with symlinks.. Anyways that is out of the scope of this PR :)

isavegas:
Yeah, I stumbled across the issue with GNUWin32's Zip port not supporting `-y` when testing the deploy, but set it aside for now. Would making `-y` platform dependent be a viable solution?

PR #866, opened 2018-01-27
Toggle Off The Record 

pago88:
Adds a button to the conversation head that can toggle the conversation's history status. It will display whether the conversation's history is currently on or off. 
When an otr modification event is sent/received, it will display a little banner in the conversation displaying who changed history on or off. 
Added some styling to change the background color of the the conversation bubbles when history is off (to mimic gmail/hangouts behavior).

### DIFF
--- a/src/locales/en.json.template
+++ b/src/locales/en.json.template
@@ -38,10 +38,15 @@
         "delete_confirm": "Really delete conversation?",
         "leave": "Leave conversation",
         "leave_confirm": "Really leave conversation?",
+        "history_on": "History has been turned on.",
+        "history_off": "History has been turned off.",
         "search": "Search people",
         "new_message": "New message received",
         "new_attachment": "New message received (image or video)",
-        "open_link": "Opening the link in the browser..."
+        "open_link": "Opening the link in the browser...",
+        "toggle_history":"Toggle conversation history",
+        "toggle_history_on":"History is on",
+        "toggle_history_off":"History is off"
     },
     "notification": {
         "one": "Notification",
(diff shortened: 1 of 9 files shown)


averissimo:
Great work on this!! I really like the hangouts-style notification of the otr status!! 👍 

There is just a scenario that I'm not 100% happy with, but this is mostly from my code and not related to yours.

1. Turn history off and icon changes accordingly
1. Icon changes back to on
1. Icon changes back to off

I think this happens when the call to `modifyotrstatus` takes some time to be performed and a `syncrecentconversation` happens in the mean time.

I'm currently on a bad internet connection, so this might be specific to me.

**Do you ever experience this?**

ps. It does not matter if in first step I am turning history on or off. Used off to follow through on use case.

pago88:
@averissimo I may have seen that at some time, but I've been trying to replicate it and I don't get that behavior. What does happen to me is that, I toggle the history button, and the amount of time it takes for the banner to show up saying I turned history on or off varies. Sometimes it's really short (almost instantaneous, but not quite), and other times it takes a few seconds. I'm sure it has to do with the time it takes to sync.

averissimo:
yeah, I was trying to work around this with ( 3c43b0b ), by calling history as soon as a callback for `modifyotrstatus` client call.

*(note: calling history and changing the way it was merged to avoid duplicate entries and out of order messages)*

But my internet connection is pretty bad now and I can't even send messages most of the time -_-

pago88:
@averissimo What do you think about something like this  pago88@b85ec5d5fe422ff511df2e884e84c9ebe1c48a94 that I was testing . It simply tells the client to sync as soon as the client successfully toggles the otr status. In my testing it takes about 1 - 2 seconds every time to receive the notification about the otr change.

PR #936, opened 2018-06-04
use tarball instead of git to resolve electron packager issue 

leboff:
electron-userland/electron-packager#796


davibe:
does this mean that current build is broken ?

PR #1004, opened 2018-11-07
Fix popup showing wrong text when opening links.

HomerSp:
This fixes a minor bug in a popup that shows when you would open links that said 'About YakYak' instead of 'Opening in browser'.


averissimo:
Thanks for the two patches @HomerSp and the hangupsjs fixes

PR #1025, opened 2018-12-16
add Simplified Chinese translation

HenryHu:
Add Chinese translation. I'm not sure how to test it, so I only checked it manually.


averissimo:
Super, it will be available in the beta rolling release soon!

PR #1042, opened 2019-01-29
Changes next/prev conversation shortcuts on macOS

phongnh:
Because Cmd+Tab and Cmd+Shift+Tab are default shortcuts for switching
apps on macOS and cannot be used to change next/prev conversation.

### DIFF
--- a/src/ui/views/menu.coffee
+++ b/src/ui/views/menu.coffee
@@ -23,8 +23,8 @@ acceleratorMap = {
     quit: { default: 'CmdOrCtrl+Q' }
     zoomin: { default: 'CmdOrCtrl+Plus' }
     # Platform specific
-    previousconversation: { default: 'Ctrl+K', darwin:'Command+Shift+Tab' }
-    nextconversation:  { default: 'Control+J', darwin:'Command+Tab' }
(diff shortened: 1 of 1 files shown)


averissimo:
Seems logical! I cannot test this as I don't run MacOSX.

jmooo:
Any reason `cmd+k` and `cmd+j` were chosen? More standard shortcuts on macOS used for say, Chrome and Terminal etc, are `control+tab`/`control+shift+tab` and duplicated via `cmd+shift+[`/`cmd+shift+]` (those are open and close bracket keys)

Seems like the original settings could have just been moved to `control` instead of `command` which would have fixed it in a more standard way

image
image

mkellerman:
I agree with @jmooo. Sticking with Mac convention would be best. CTRL+SHIFT+TAB and CTLR+TAB.

This is the same for Safari and Chrome.

averissimo:
Going with a change to Ctrl+Tab as it seems it's the standard in macOS

If this displeases part of the user base please discuss and together reach a conclusion

*(maybe Electron supports multiple shortcuts for the same menu actions)*

PR #1071, opened 2019-08-13
Fix the Unicode mapping for 💩

michaelherold:
This mapping was incorrectly set to [flexed biceps][1] instead of
[poop][2] which made it very confusing when trying to type it out.

[1]: https://www.fileformat.info/info/unicode/char/1f4aa/index.htm
[2]: https://www.fileformat.info/info/unicode/char/1f4a9/index.htm


averissimo:
<3

Issue #1095, opened 2020-01-06
No new releases?


mkellerman:
Seems like the last release dates of Nov 8, 2018. What is the status on this project? PR's are pilling up. Is this project dead?

joshlewis:
As of today, the app (or Google?) forced me to sign out of YakYak, and I can't sign back in because the built-in "browser" is considered to be insecure by Google. So I can't use the app anymore. 👎

mkellerman:
> As of today, the app (or Google?) forced me to sign out of YakYak, and I can't sign back in because the built-in "browser" is considered to be insecure by Google. So I can't use the app anymore. 👎

Check for a fix here: https://github.com/yakyak/yakyak/issues/1087#issuecomment-565170640

Hence why a new release would be nice, that includes the fix...

joshuacant:
Seems like the maintainer has moved on, it happens. Has anyone forked this yet and incorporated all the changes from the last couple years into a new build?

mkellerman:
I can help setup a new fork, and create a build in Azure and such.. but I’m no dev. I would be useless to approve/comment on PRs.

Would be nice if the maintainer could give us a status update. I’ve seen recent commits being done, but no releases.

joshuacant:
The most recent commits are from Jan/Feb of 2019, which is basically a year ago, so it has been quite a while since there was activity. I don't mean any of this as an attack on the current maintainer. Things happen, interests change, time moves on. It probably didn't help motivation any when Google cancelled and then un-cancelled Hangouts in 2019.

I think someone taking the project on would need to be able to manage incoming PRs and doing builds for MacOS, Linux, and Windows. Either that, or split the duties between a couple people.

atomicwrites:
I thought it said he had stopped, maintaining it since hangouts was deprecated, but i don't see it now.

averissimo:
Sorry about that, but I'm no longer actively maintaining yakyak.

Both me and the previous developer stopped using hangouts and if someone wants to step up I gladly give him permissions.. *(or ask the previous dev to give it)*

I'm merging pending PR requests and will be updating the internal packages it uses.

I'm hoping to release a new version today.

This is all I can do.

averissimo:
New release is available at [v1.5.4-beta-rolling](https://github.com/yakyak/yakyak/releases/tag/v1.5.4-beta-rolling)

I'd like if someone could test it for a bit to see if everything works fine.

I'll wait until users in Windows and Mac say that it's working for them.

[...]

PR #1107, opened 2020-01-27
Fix login.coffee

ryantiger658:
Fixing the startup loop. 

However, after login I am hanging on `One moment please...` message with no obvious errors in the log. It seems to be related to the URL that we get back from Google, but I can't tell for sure. 

See https://github.com/yakyak/yakyak/issues/1095#issuecomment-578852932


averissimo:
Do you still have this on the v1.5.4-beta-rolling? *(downloaded in the last day)*

This should be fixed with a more recent version of Electron *(which, in turn uses a different user agent)*

averissimo:
Nevermind, I read your post in the "No new releases" issue

ryantiger658:
Yes I was having it on v1.5.4-beta-rolling as of 10am this morning. If I was reading it right that release is built off of master and master has the broken code.

averissimo:
Ok, I was under the impression that it was solved with an update to electron, but I was completely wrong.

Seems like Google doesn't like the default user-agent and finds it unsecure.

I manually merged your commit, I'm closing the issue.

averissimo:
Thanks!

averissimo:
There's still a problem with the cookie not being retrieved.

Once I login it doesn't close the window and proceeds to open yakyak.

Working on it now

Issue #620, opened 2017-03-09
Option to disable Inline Pictures


cron410:
This feature would be very useful for when using Yakyak in the workplace.

cron410:
Closed due to inactivity?

averissimo:
yes, the bot went crazy today :)

cron410:
+1

averissimo:
Can you be a little more descriptive on what you are asking?

Stoppablemurph:
I think the requester (and me) would like an option to not show images received by default. Some sort of show/hide functionality (with a configurable default state) would be perfect. It's kind of awkward when I'm at work and my sister starts spamming baby pictures in our family's group chat.

averissimo:
This can be probably achieved by adding a new option and changing the code in messages.coffee

```
# image formatter
    (seg) ->
        href = seg?.link_data?.link_target
        imageUrl = getImageUrl href # false if can't find one
        if imageUrl and preload imageUrl
            div ->
                img src: imageUrl
```

with a if clause that checks the option and instead of outputing a `img` tag it outputs a link tag saying it is an image

PR #1108, opened 2020-01-28
Add show-hide preview for attachment images too

AlessSala:
This is a feature #620 follow-up.
A new i18n string is added: 'no_preview_image_click_to_open'.

### DIFF
--- a/src/locales/bn.json
+++ b/src/locales/bn.json
@@ -41,7 +41,8 @@
     "search": "লোকেদের অনুসন্ধান করুন",
     "new_message": "নতুন বার্তা",
     "new_attachment": "New message received (image or video)",
-    "open_link": "Opening the link in the browser..."
+    "open_link": "Opening the link in the browser...",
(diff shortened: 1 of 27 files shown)


averissimo:
I'm not really a fan of "Hidden" in the message "Hidden image: click to open it in the browser"

Something a bit more explanatory as: "Image preview is disabled: click to open it in the browser"

What do you think?

AlessSala:
Sounds good to me, right you are. 
What's the plan, then? Have i to delete, fix and re-pull?

averissimo:
Just commit it again and the pull request is updated... You don't need to do anything here :)

AlessSala:
Lol i'm sorry for the nooby question :p

averissimo:
Don't be afrraid of asking questions!!

I'm more than happy to share what I know with anyone that is curious. I believe most of the community here in Github feels the same :smile:

PR #1111, opened 2020-01-30
add arabic language

belaassal:
transalate All strings to arabic language.


averissimo:
Thanks for your contribution.

PR #1120, opened 2020-02-04
Work around window position not always being restored properly

HomerSp:
This fixes an issue where the window position isn't always restored properly. For example, I'm running Linux and have a panel at the top of my primary monitor, but not on my secondary. If I move yakyak to the top of my secondary monitor, the next time I open it up, it won't be restored at the top of the secondary monitor, but will instead be a few pixels down from the top, corresponding to the bottom of the top bar on my primary monitor.
The fix actually allows the system to handle the positioning, so the part where it checks if it's within the bounds of the monitors and such isn't necessary anymore.

I haven't noticed any issues myself, and it seems to work pretty much identical to how it did before, without the overhead of the manual checking.

Do let me know if you want any changes to be done, or if more testing is required.

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -7,6 +7,7 @@ path      = require 'path'
 tmp       = require 'tmp'
 session   = require('electron').session
 log       = require('bog');
+windowState = require('./ui/models/windowstate')
 
 [drive, path_parts...] = path.normalize(__dirname).split(path.sep)
 global.YAKYAK_ROOT_DIR = [drive, path_parts.map(encodeURIComponent)...].join('/')
@@ -121,14 +122,18 @@ app.on 'ready', ->
                 process.env[t.env] ?= if purl then "http://#{purl}" else ""
                 rs()
 
+    bounds = windowState.getBounds()
+
     icon_name = if process.platform is 'win32'
         'icon@2.png'
     else
         'icon@32.png'
     # Create the browser window.
     mainWindow = new BrowserWindow {
-        width: 730
-        height: 590
+        x: bounds.x
+        y: bounds.y
+        width: bounds.width
+        height: bounds.height
         "min-width": 620
         "min-height": 420
         icon: path.join __dirname, 'icons', icon_name
--- a/src/ui/models/windowstate.coffee
+++ b/src/ui/models/windowstate.coffee
@@ -1,0 +1,19 @@
+Client = require 'hangupsjs'
+settings = require('node-persist')
+
+{throttle, later, tryparse, autoLauncher} = require '../util'
+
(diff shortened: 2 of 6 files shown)


src/main.coffee
```diff
@@ -7,6 +7,7 @@ path      = require 'path'
 tmp       = require 'tmp'
 session   = require('electron').session
 log       = require('bog');
+windowState = require('./ui/models/windowstate')
```
averissimo: This breaks the separation we have between main.coffee and UI

I'd prefer if we used the communication between the main process and UI instead of sharing code.

Maybe use the event 'ready-to-show' on the UI to reposition the window? Or whatever infrastructure was used befoer

src/ui/models/windowstate.coffee
```diff
@@ -0,0 +1,19 @@
+Client = require 'hangupsjs'
+settings = require('node-persist')
```
averissimo: Why would it be better to use node-persist instead of `localStorage`

src/ui/views/controller.coffee
```diff
@@ -42,68 +42,6 @@ handle 'update:viewstate', ->
     setLeftSize viewstate.leftSize
     setConvMin viewstate.showConvMin
     if viewstate.state == viewstate.STATE_STARTUP
-        if Array.isArray viewstate.size
```
averissimo: How does this work when you close Yakyak on a 2 monitor setup and then open it again on a 1 monitor setup?

I can't test this scenario right now.

This code was put in place because of that *(see the comments)*

averissimo:
I've just tested this under Ubuntu 20.04 *(gnome 3.34.3)* in a single monitor environment.

Test:

* Position Yakyak window in top right corner
* Quit Yakyak
* Start again

Expected: Window in same position

What I got: Seemed like window was moved a bit down *(distance may be the titlebar height)*

src/ui/models/windowstate.coffee
HomerSp: localStorage requires an active BrowserWindow, but since this code is used to initialise the position of the BrowserWindow there's no way of actually using that. That's why node-persist is required.

HomerSp:
You're right, seems like there's a bug on linux with electron that causes it to move down. Have a look here: https://github.com/electron/electron/issues/10388

src/ui/views/controller.coffee
HomerSp: Just tested this, and it opens just fine at the right side on my primary monitor if I move it to my secondary (it's to the right of my primary), close it, and disconnect the secondary monitor. So it looks like this is working at least.

PR #1121, opened 2020-02-05
[do-not-merge] React transition? 

averissimo:
Proof of concept to improve yakyak infra-structure.

* Implements About page in react

Questions:

* Is this the best way to migrate view by view?
    * Have ReactDOM render and unmount everytime?
    * Once all views are implemented as component, how to best handle routing?


averissimo:
I hope this can sparkle a discussion and that a someone gets motivated to work on this

davibe:
just to add some noise. Thinking about following the trends one may think about leaving coffee for typescript (Although I still secretly love coffeescript)

averissimo:
@davibe Coffeescript code is beatautiful! Or maybe I'm bias for all the code I did in yakyak.

I've started trying to migrate hangupsjs to coffeescript v2 and haven't had any problems yet... Have you migrated any projects to that version and have some tips on what to look for?

I'm a bit cautious on pulling that trigger lol

HomerSp:
Personally I think this would be a good idea - trifl is deprecated and hasn't been updated in a long time. It will be a bit of work to transfer everything over to react, but I think it would be worth it.

averissimo:
The codebase needs to be updated on the long term, the coffeescript update was probably the first step.

This pull request is just a proof of concept that works alongside trifl to ease the transition by being able to first convert the views and then change the backend. It's a conservative approach.

But again, it's a proof of concept, if you have other ideas please move on them :-) I'm not really a javascript developer, so my knowledge and good practices are a bit limited.

averissimo:
Rebased the changes on top of current codebase _(without remote)_

It seems to work as it was working before, I corrected an unreadTotal bug

HomerSp:
I finally managed to look through this and I feel like it's a good starting off point. Perhaps we should set up a separate branch for the react transition so we can work on that simultaneously with other fixes and such?

As for the code itself, we should probably use a component manager or such to handle creating react components, rendering them and unmounting/destroying them.

PR #1126, opened 2020-02-13
hiding window with ctrl+w

JanBN:
Hi, I'm using ctrl+w to close window the same as in the web browser. Here is pull request adding ctrl+w support.


averissimo:
Thanks, merged!

averissimo:
Wait I'm going to revert the merge unless you can add a sanity check.

* If tray isn't showing then quit the app
* Whatever applies in a macOS

The reasoning behind is that there's no method *(known to me)* to re-open the window if it's hidden and no tray icon is present *(in Ubuntu/Linux)*

PR #1157, opened 2020-05-06
fix again the RTL problem

chaim-chv:
i assume this is the correct way to do.
please release it again
and i will test it
thank you again. i apriciate that!

### DIFF
--- a/src/ui/views/messages.coffee
+++ b/src/ui/views/messages.coffee
@@ -246,8 +246,7 @@ drawMessage = (e, entity) ->
     mclz = ['message']
     mclz.push c for c in MESSAGE_CLASSES when e[c]?
     title = if e.timestamp then moment(e.timestamp / 1000).calendar() else null
-    div id:e.event_id, key:e.event_id, class:mclz.join(' '), title:title,
-        dir: 'auto', ->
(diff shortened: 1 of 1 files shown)


src/ui/views/messages.coffee
```diff
@@ -246,8 +246,7 @@ drawMessage = (e, entity) ->
     mclz = ['message']
     mclz.push c for c in MESSAGE_CLASSES when e[c]?
     title = if e.timestamp then moment(e.timestamp / 1000).calendar() else null
-    div id:e.event_id, key:e.event_id, class:mclz.join(' '), title:title,
-        dir: 'auto', ->
+    div id:e.event_id, key:e.event_id, class:mclz.join(' '), title:title, dir: 'auto' ->
```
averissimo, line 249: This is not correct, the ` -> ` is an argument for `div` function

You can build it on your own computer by installing [nodejs](https://nodejs.org/en/)

Then it's as easy as:

* download and extract the repository to a folder
* running in the command line *(inside de folder)*:

```
npm install
npm start
``

chaim-chv: ok i'll try it
thank you

PR #1187, opened 2020-06-21
Updated read icons for colorblind people

### LINKED ISSUE #1184: Colorblind problem
Hi all,

I'm colorblind (deuteranomaly) and I can't distinguish the tray icon colors for read and unread states so normally I locally change the icons making the read state icon have a blue tone and leaving the unread as it is.

It could be any possiblity to make that change to be permanent in the repo so I don't need to make the change with every release version.

Thanks in advance.

_PS: If you accept the proposition I can make a PR with my modified icons as read state icons_

kix2902:
Updated read icons with a blue overlay
Resolves #1184


averissimo:
Thanks for the PR. It's already building on the 1.5.8-beta-rolling and should be out in under 20mins.

This improvement is going to fast track that release.

Would it also be easily readable if it had a white silhouette and transparent background?

For instance: https://user-images.githubusercontent.com/211358/85289082-29a59400-b497-11ea-997a-99876c862950.png

ElSenor64:
> For instance: https://user-images.githubusercontent.com/211358/85289082-29a59400-b497-11ea-997a-99876c862950.png
I would be quite happy with this icon. I do support this feature very much, but would also appreciate it if this were a setting that you could change; now YakYak's icon sort of sticks out in my tray...
image

kix2902:
I've made a test in my computer and the white-border icon is different enough to the green icon so it could be used for the "read" state without problem

averissimo:
Yakyak is currently building v1.5.10 that should have a different default icon for "read" status. _(the one I linked in a previous comment)_

Nevertheless I've added an option _(defaults to false)_ to further enhance the tray icon to the blue one under _"View -> Tray -> Use colorblind tray icon"_

We're **not** going back to the grey icon, which I think everyone will agree.

PR #1196, opened 2020-06-30
Fix frame/decoration not showing on linux.

### LINKED ISSUE #1194: No window decoration on yakyak 1.5.8
Not sure if that is by design, but there are no window decorations (title bar, min/max/close buttons etc). Escape key also doesn't seem to hide yakyak anymore, perhaps because of the missing window focus. On ubuntu 20.04.

HomerSp:
This fixes #1194

### DIFF
--- a/src/main.coffee
+++ b/src/main.coffee
@@ -150,6 +150,8 @@ app.on 'ready', ->
 
     if process.platform is 'darwin'
         windowOpts.titleBarStile = 'hiddenInset'
+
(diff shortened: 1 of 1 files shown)


HomerSp:
I think the problem is that the code evaluates to (void 0) in the else case which, it seems, electron now treats as false.

averissimo:
I've merged this manually, thanks for the quick pull request

PR #1197, opened 2020-07-01
Fix macOS titlebar issue

jmarquis:
1.5.9 introduced a regression where the app titlebar/window buttons are no longer inset like they should be on macOS (see screenshots). This PR fixes the innocent typo that's responsible 😄 

### Before:
before

### After:
after


averissimo:
Oopsie :)

thanks!

HomerSp:
Whoops, sorry about that!

PR #1207, opened 2020-07-06
Version start

averissimo:
@HomerSp does this make sense? _(is it well implemented?)_

I just asked a user for a screenshot and I figured out that having the version would help alot in the future, along with hangusjs'.

This assumes we also add the following to `hangupsjs`

```
Client.VERSION            = JSON.parse(fs.readFileSync(syspath.normalize syspath.join(__dirname, '..', 'package.json'))).version
```


HomerSp:
Looks good to me! Definitely a good idea to show the version in stdout - I'm actually surprised this wasn't in there already :D

PR #1209, opened 2020-07-07
adds homersp to list of contributors in readme

averissimo:
@HomerSp is it ok to add you to the list of contributors?


HomerSp:
Of course! :)

PR #1233, opened 2020-09-08
fi.json

survivor303:
Updated strings and some other cleaning.


averissimo:
Thanks, I commited it manually as there was some issues with commas.

Please check if everything is correct on v1.5.10

PR #1273, opened 2020-12-10
Fix endless scrolling when scrolling through chat history

HomerSp:
This fixes an issue in v1.5.10 where scrolling up through the chat history would cause it to endlessly load older messages.
The problem happens because, I'm guessing, in electron 11 an element that has flex-flow as column-reverse will also reverse the scrollTop value - meaning it starts at 0 when not scrolled and then it goes down into the negatives as you start scrolling.
This change also means that we no longer have to calculate the position of the top message when adding older messages to the view, since it's based on the bottom of the view. So I changed this around to only save the scrollTop and then set that back when we added the older messages. For whatever reason the scroll position still changes when adding items, so we do have to save it. I also had to set the saved scroll top position at a later time as just setting it in adjustMainPos meant that it would be overridden.

### DIFF
--- a/src/ui/views/applayout.coffee
+++ b/src/ui/views/applayout.coffee
@@ -27,12 +27,20 @@ onScroll = throttle 20, (ev) ->
     el = ev.target
     child = el.children[0]
 
+    # use the saved scroll value if we have one
+    if el.hasAttribute('scrolltop')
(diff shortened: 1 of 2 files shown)


averissimo:
I'm trying to contact the developer to grant you permissions, sorry about the delay

Issue #1272, opened 2020-12-09
Download for v1.5.10 Windows 64bit is invalid


HayesSolutions:
Hello

In case you are unaware, the yakyak-1.5.10-win32-x64.zip file is invalid and Windows cannot open the folder.

Thanks and kind regards

Dominic

HomerSp:
Duplicate of #1269

averissimo:
@HayesSolutions can you test current v1.5.11 pre-release?

Please don't mind it being called v1.5.10 on the file name, I'm sorting that out before making it an official release

https://github.com/yakyak/yakyak/releases/tag/v1.5.11

HayesSolutions:
> 
> 
> @HayesSolutions can you test current v1.5.11 pre-release?
> 
> Please don't mind it being called v1.5.10 on the file name, I'm sorting that out before making it an official release
> 
> https://github.com/yakyak/yakyak/releases/tag/v1.5.11

Thank you for following this up.  Unfortunately, ths produces the following error
![image](https://user-images.githubusercontent.com/57859813/101927086-fa611b80-3bcb-11eb-91de-51cdd9bcc299.png)
Kind regards
Dominic

averissimo:
Oh man. That's bad, I'll check out on my old windows pc.

Has the problem with the file being invalid been corrected?

HayesSolutions:
> 
> 
> Oh man. That's bad, I'll check out on my old windows pc.
> 
> Has the problem with the file being invalid been corrected?

Hello
No worries, these things happen.  I don't think the file has been corrected yet and I have reverted to 1.5.9 for the time being.  I'd like to update as this version does not always show when I have unread messages and that causes me problems at times!
Best regards
Dominic

HomerSp:
@averissimo The file looks to be generated as a zip for me now, so that should be working at least.

@HayesSolutions Any chance you can post the output from the console (ctrl+alt+i)? Would help us knowing what's wrong. I'm not able to reproduce it myself.

HayesSolutions:
> 
> 
> @averissimo The file looks to be generated as a zip for me now, so that should be working at least.
> 
> @HayesSolutions Any chance you can post the output from the console (ctrl+alt+i)? Would help us knowing what's wrong. I'm not able to reproduce it myself.

Morning HomerSp

In 1.5.9 the console log is present, see below, but after replacing with 1.5.10 the error occurs before the app has loaded and there is no console available.
1.5.9 Console
[...]

PR #1274, opened 2020-12-10
closes #1269 and #1272

### LINKED ISSUE #1269: Downloadable artifact yakyak-1.5.10-win32-x64.zip is actually a tar
Expected format is .zip, as it was in previous releases

averissimo:


averissimo:
The changes 5 months ago from 9c43f228bec7a010993a4997139caa3b3daa3e22 were meant to use tar.exe utility in windows.

But that should be specific to that platform alone

Issue #1242, opened 2020-09-23
Presence and "… is typing" 


jcvignes:
Hello,

I love YY and I use it daily. For some reason I don't see the "green dot" when my friend is online, nor does it tell me when she is typing… She just sent me a screenshot (using YY too) and she sees a green dot next to me and whether I'm typing…  I thought it might be a Hangout issue, but on the web app (Hangout) I do see the green dot next to her name? 

I know it's a little thing but it's driving me crazy so any help would be greatly appreciated!

Thanks,
JC

asacarny:
Seconding this issue (and I think YakYak is great as well!) I do not see any green dots on my contact list any more.

HomerSp:
Can you check the beta build to see if it's working as intended there?

PR #1275, opened 2020-12-11
Fix the presence indicator not working anymore

HomerSp:
As can be seen in eg #1242 the presence indicator does not work currently. This is caused by a few different issues, for one they did some server-side changes to some object names, causing the initial presence loading to not work. They also completely removed the presence value from the server event, and it is instead replaced by a null. I fixed this by instead relying on the timestamp in that message, which will be checked against the self presence timestamp and if it's older than an offset (15 minutes currently because that seems to be what it is in the web app) the user will be marked as not present.


averissimo:
Looks really good!

PR #1276, opened 2020-12-11
Apple Silicon build configuration

michal-olszewski:
Added build configuration for darwin with arm64 architecture which corresponds to native binary for Apple Silicon based Macs.


averissimo:
Thanks for the change, it's now building on travis for version v1.5.11 right now before releasing it

PR #1286, opened 2020-12-19
Added enableRemoteModule to webPreferences

tthordarson:
Added enableRemoteModule to webPreferences. This fixes subsequent issues when remote is imported from electron.


HomerSp:
Thanks for the PR! I'm in the process of converting the remote code to ipc, so I'm hoping that we'll be able to do without this soon as electron are completely deprecating remote in the near future from what I've read.

PR #1290, opened 2020-12-22
Convert remote calls to ipc

HomerSp:
Should be fairly self-explanatory, but this basically removes the dependency on the remote module, and instead uses ipc calls. It's a bit of a wip, but everything I've tested so far seems to work.

One annoyance is that functions can't be sent through ipc (for obvious reasons), so I had to change the menu handlers to use actions instead of doing things directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     ]
   },
   "dependencies": {
+    "@electron/remote": "^1.0.0",
     "auto-launch": "^5.0.5",
     "autosize": "^4.0.2",
     "bog": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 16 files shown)


averissimo:
Awesome rewrite!

This will make sure future electron versions are supported.

I saw that you've kept the `menuaction` method. Seems valid as it keeps using the action layer, which justifies the overhead of messages back and forth.

HomerSp:
Thanks for testing! It's also working fine here, so I might just merge it so we can get hopefully get more people testing it.

Yeah, I was considering writing some extra code for the actions that are supposed to go to just the main thread. But I didn't really think it was worth the extra effort since the calls only happen when you press the menu item, and the very slight overhead adds maybe a few milliseconds at most for the action (untested, but I'd imagine). If it turns into a problem I'll rewrite it though.

PR #1293, opened 2020-12-22
Update nodejs to v15

HomerSp:
Now that the deploy issue has been fixed we should be safe to update node.js to v15.5.0. This also requires updating electron.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 14
+  - 15
 notifications:
   email: false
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 14
+  - 15
```
averissimo, line 3: you can use `node` instead of a specific version to use the latest

https://docs.travis-ci.com/user/languages/javascript-with-nodejs

HomerSp: Thanks, I did consider doing that, but thought that it would be better to target v15 specifically in case the eventual v16 brings in any breaking changes that would require us to, temporarily, downgrade to 15. I'll update it in the future though!

PR #1311, opened 2021-03-04
Use the new getvideoinformation from hangupsjs

HomerSp:
We can now properly get public urls for video attachments, so use this new function. This also updates images to use the proper thumbnail and public url.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5993,7 +5993,7 @@
       }
     },
     "node_modules/hangupsjs": {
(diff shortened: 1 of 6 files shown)


averissimo:
So I guess hangouts api won't be shutting down anytime soon :-)

Great catch (on the protocol)!

HomerSp:
I've known about the video information endpoint for a while now, but I wasn't able to get it working - I would always get authentication failed or somesuch. The key to getting it to work was figuring out that the origin URL used to generate the sapisidhash was wrong. Once I did that it's now working properly! :)
I'll probably add a loading indicator or something when it's retrieving the public URL, but for now it's good enough.

PR #1393, opened 2022-01-06
Created zh_TW.json

olivertzeng:
BTW, WDYM by pop and gruvy theme? And those themes doesn't even work.


olivertzeng:
Aha! All themes work except the pop theme. I'll put this into issues.

